PR #7, opened 2020-03-24
Add root navigation and home page layouts

austin94:
## Description

This PR includes the first UI and page layouts for the new Garden site. It includes:

- Root navigation including headers/footers
- Layouts associated with the home page

### Out of scope

This PR doesn't include any animation, production ready image/svg assets, final copy, or other fine-tuning. Please limit feedback to the overall layout and mobile-responsiveness. As we include more layouts we will need to update the hover/focus/active/selected states for all anchors as well.

We also don't have a complete idea on how we will source content for the 'Site Updates` section. This is not included at this time.

### Layouts

This PR introduces 2 common layouts for use later on:

- `MaxWidthLayout`
  - Applies a consistent max-width to content. Will be used in the Home page and 2-column layouts.
- `RootLayout`
  - Positions content with header and footer components

### Images

We are using 3 types of progressive image loading:

- The Garden "tombstone" in the search row of the home page is faded-in once loaded. Since this is above-the-fold we aren't using a base64 approximation for this image.
- The Foundation images apply an SVG trace approximation until the assets are loaded.
- The Patterns image uses a base64 approximation as it is below-the-fold.

You can view these different image states on initial page load.

### Garden News

I've created a local `garden-news-source` plugin to source articles from a `content/news.yml` file. 

@talia-eisenberg and the other designers. You can click the `View deployment` link below to view the most up-to-date view of this PR.

PR #9, opened 2020-03-25
chore: update base config and documentation

jzempel:
Will hold in draft until #7 is merged.


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+   you apply an appropriate squashed [conventional
+   commit](https://conventionalcommits.org/) message when it's time to
+   merge to master.
+1. When your branch is ready, open a new pull request via GitHub.
+   The repo [PR template](PULL_REQUEST_TEMPLATE.md) will guide you
+   toward describing your contribution in a format that is ultimately
+   suitable for a structured conventional commit.
+1. Every PR must pass CI checks and receive at least one :+1: to be
+   considered for merge.
+1. Garden
```
hzhu: It seems like this is a numbered list with `1.`'s. Should this list be ordered from 1 - 6?

jzempel: @hzhu markdown takes care of that for us :wink: See https://raw.githubusercontent.com/zendeskgarden/react-components/master/.github/CONTRIBUTING.md

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+request, follow the [steps](#pull-request-workflow) below. We follow a
+[code of conduct](CODE_OF_CONDUCT.md) as our guide for community
+behavior.
+
+## Development Workflow
+
+Before you start, be sure [yarn](https://yarnpkg.com/en/) is installed
+on your system. After you clone this repo, run `yarn` to install
+dependencies needed for development. After installation, the following
+commands are available:
```
austin94: Just double-checking here. Once we enable the Abstract layer sourcing it will be impossible for anyone outside of the Zendesk org to build the site locally. Should we call that out here?

package.json
```diff
@@ -1,18 +1,19 @@
...
-    "lint:css": "stylelint \"src/**/*.{js,ts,tsx}\"",
+    "lint:css": "stylelint 'src/**/*.{js,ts,tsx}'",
     "lint:js": "eslint ./ --ext js,ts,tsx --max-warnings 0",
     "lint:md": "markdownlint ./*.md src/**/*.md",
-    "serve": "gatsby serve",
-    "start": "gatsby develop",
-    "test": "echo \"Write tests! -> https://gatsby.dev/unit-testing\" && exit 1"
+    "prepare": "yarn build",
+    "start": "gatsby develop --open",
+    "test": "prettier-package-json --list-different && yarn format:all --check && yarn lint"
```
austin94: Would we want to include `yarn tsc` to this test script?

package.json
```diff
@@ -1,18 +1,19 @@
...
+    "format:all": "prettier --loglevel warn '**/*.{js,jsx,ts,tsx,json,md,yml}' '!.github/ISSUE_TEMPLATE.md'",
     "lint": "yarn lint:js && yarn lint:css && yarn lint:md",
-    "lint:css": "stylelint \"src/**/*.{js,ts,tsx}\"",
+    "lint:css": "stylelint 'src/**/*.{js,ts,tsx}'",
     "lint:js": "eslint ./ --ext js,ts,tsx --max-warnings 0",
     "lint:md": "markdownlint ./*.md src/**/*.md",
-    "serve": "gatsby serve",
-    "start": "gatsby develop",
-    "test": "echo \"Write tests! -> https://gatsby.dev/unit-testing\" && exit 1"
+    "prepare": "yarn build",
```
austin94: This site doesn't follow the standard pre-post script process as versioned packages. Why do we need to build the fulls site for this event?

jzempel: I actually want this repo to follow the same conventions as the other Garden repos. You run `yarn` and everything is ready to go.

package.json
jzempel: Maybe. We don't do that with the other repos (but maybe we should). I'll update.

.github/CONTRIBUTING.md
jzempel: Not at this time. Let's see how things evolve. I don't envision external involvement beyond filing issues. This is intended for the internal audience.

package.json
austin94: The prepare/build step is required in the other repos due to the lerna boostrap process.

Since build is unrelated to being able to develop the site I would lean toward it not being added here.

.github/CONTRIBUTING.md
hzhu: I noticed that some content/sentences in these markdown files could be interpreted as encouraging external involvement beyond filing issues. Allowing to see how things evolve sounds good 👍.

package.json
jzempel: I don't see it that way. Every other Garden repo (eslint-config, stylelint-config, svg-icons, etc) follows this convention. That is, once `yarn` is run, everything is intact for `yarn start`, etc. When lifecycle scripts aren't necessary, then run `yarn --ignore-scripts`.

jzempel: The approach simplifies the documentation and consistency for anyone to get up-and-started for website contribution.

hzhu: > The approach simplifies the documentation and consistency for anyone to get up-and-started for website contribution.

Should the approach be geared towards maintainers since external involvement could be limited to filing issues? Seems like it would tradeoff consistency between repositories, but maybe that might not be a bad thing since the nature of this repo is a bit different? This also seems like a topic that could be reassessed as the project evolves.

jzempel: Yes. And the maintainers for `website` include both eng and non-eng individuals. We must make the entry process super simple – and consistent with the other repos.

PR #22, opened 2020-04-22
feat: introduce Garden elements to markdown rendering

austin94:
## Description

This PR adds some general visual updates to the site and adds Garden elements to our markdown renderer.

## Detail

- Abstract Layer Remark Sourcing Plugin
  - Add support for using native `` tags in HTML and JSX
  - This requires parsing node string content with `cheerio` to access nested attributes correctly
- Update code example footers to `grey-200`
- Update footer icon size and color to match new designs
- Update home Anchor usages to match default styling
- Ensure TOC headers match styling of sidebar section headers
- Add the following markdown elements with their respective Garden elements
  - `h2-h6`, `p`, `blockquote`, `hr`, `inlinecode`, `pre`, `code`, `ul`, `ol`, `li`, `table`
  - Both tables and lists required some context usage to ensure our elements are rendered correctly.
  - The code block examples use the same theme as our typescript code examples and support dynamic language tags.

## Out of Scope

Vertical spacing and typography still have some work. Once the `<Paragraph>` element is added that can help influence our spacing requirements.

## Demo

I've added some extra markdown styles to the `/content/basics` page for you to view.

## Checklist

- [ ] ~:black_nib: copy updates are approved (add the content strategist as a reviewer)~


src/components/MarkdownProvider/components/Typography.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+`;
+
+export const StyledBlockquote = styled.blockquote`
+  margin-top: ${p => p.theme.space.md};
+  margin-bottom: ${p => p.theme.space.md};
+  border-left: ${p => p.theme.borders.sm} ${p => getColor('grey', 300, p.theme)};
+  padding-left: ${p => p.theme.space.sm};
+  color: ${p => getColor('grey', 600, p.theme)};
+  font-style: italic;
```
hzhu, line 58: It was mentioned in one of my PRs to avoid using italics in Garden due to i18n issues. Would that apply here too?

src/components/MarkdownProvider/components/Typography.tsx
```diff
@@ -0,0 +1,77 @@
...
+  margin: ${p => p.theme.space.md} 0;
+  border-top: ${p => p.theme.borders.sm} ${p => getColor('grey', 300, p.theme)};
+`;
+
+/**
+ * The SyntaxHighlighter component provides it's own `<pre>` tag.
+ * This ensures valid DOM nesting.
+ */
+export const StyledPre: React.FC = ({ children }) => {
+  return <>{children}</>;
```
hzhu, line 76: Is this React fragment needed because additional elements will be added in the future? Could the component return `{children}` alone?

austin94: Since children are optional this fails the TS type check. This ensures that a React node is always returned.

src/components/MarkdownProvider/components/Typography.tsx
austin94: This is specifically for the "emphasis" usage which is a valid markdown element. If we were to localize this content then the markdown source could not include that element if necessary.

content/uxcs/basics.mdx
```diff
@@ -18,6 +18,23 @@ Zendesk doesn’t speak jargon. Use Plain Language instead. Jargon is confusing,
...
+We now support a _wide_ variety of **styles** and `inline code` syntaxes.
+
+1. Ordered lists
+1. Are now
+1. Supported
+
+> Including long-form block quotes
+
+- As well as unordered
+- Lists
```
jzempel, line 30: @talia-eisenberg please note that Garden has the ability to override the bullet style for unordered lists. (See https://garden.zendesk.com/react-components/typography/#lists.) We fallback to the browser default, but we can specify it to be `disc` which is probably the more typical treatment.

src/components/MarkdownProvider/components/Typography.tsx
jzempel: @hzhu the design decision was to use `italic` in this specific case. Use of italic on the product side continues to be a no-no :wink:

src/components/MarkdownProvider/index.tsx
```diff
@@ -48,12 +62,26 @@ export const MarkdownProvider: React.FC = ({ children }) => (
...
+        code: MDSyntaxHighlighter,
+        ul: UL,
+        ol: OL,
+        li: LI,
+        table: Table,
+        thead: THead,
+        tbody: TBody,
+        tr: TR,
+        td: TD,
+        th: TH
```
jzempel, line 84: Please update to map `strong` to a styled component that renders `<b>` – ideally using the new typography [isBold](https://garden.zendesk.com/react-components/typography/#md) styling.

Also, update to map `em` to a styled component that renders `<i>`. Garden opinion is that `<b>` and `<i>` are the correct elements for stylistic rendering. Since we can't infer semantics from markdown, we'll pass on  the `<strong>` and `<em>` tags.

The most recent v8.0.0 version of `css-bedrock` bears this out.

src/components/MarkdownProvider/components/Code.tsx
```diff
@@ -0,0 +1,36 @@
...
+import ReactSyntaxHighlighter from 'react-syntax-highlighter';
+import dracula from 'react-syntax-highlighter/dist/esm/styles/hljs/dracula';
+
+const StyledPre = styled.pre`
+  margin: ${p => p.theme.space.md} 0;
+  border-radius: ${p => p.theme.borderRadii.md};
+  /* stylelint-disable-next-line declaration-no-important */
+  padding: ${p => p.theme.space.sm} !important;
+  overflow: auto;
+`;
```
jzempel, line 19: Set font family stack to `p.theme.fonts.mono`.

src/components/MarkdownProvider/index.tsx
```diff
@@ -48,12 +62,26 @@ export const MarkdownProvider: React.FC = ({ children }) => (
...
-        h4: H4,
-        h5: H5,
-        h6: H6,
-        p: P
+        h2: StyledH2,
+        h3: StyledH3,
+        h4: StyledH4,
+        h5: StyledH5,
+        h6: StyledH6,
+        p: StyledP,
```
jzempel, line 70: Use the new Garden [`Paragraph`](https://garden.zendesk.com/react-components/typography/#paragraph-1) component (@hzhu will release today).

src/layouts/Titled/components/TOC.tsx
```diff
@@ -112,16 +114,19 @@ export const TOC: React.FC<{ data: IHeading[] }> = ({ data }) => {
...
-        <SM
+        <div
           css={css`
             margin-left: -${p => p.theme.borderWidths.sm};
             text-transform: uppercase;
+            line-height: ${p => p.theme.lineHeights.sm};
             letter-spacing: 0.5px;
             color: ${p => getColor('grey', 600, p.theme)};
+            font-size: ${p => p.theme.fontSizes.xs};
+            font-weight: ${p => p.theme.fontWeights.semibold};
```
jzempel, line 125: Can you abstract out this common code repeated from above?

src/components/MarkdownProvider/components/Code.tsx
austin94: This is already set with our `css-bedrock` reset. What is the reason for us to override this locally?

src/components/MarkdownProvider/index.tsx
austin94: Updated now that it is released.

src/components/MarkdownProvider/index.tsx
austin94: After discussing yesterday with design, I'm not sure this is correct for our specific usages.

Within a component boundary, usage of `<strong>` and `<em>` would be invalid as the visualization is only stylistic. But, our markdown provider, and for any long-form text within a product, the `<strong>` and `<em>` tags are required to denote differences in text for a screen reader.

Additionally, markdown doesn't provide styling, it only provides semantics. The markdown schema denotes these as [emphasis](https://github.com/syntax-tree/mdast#emphasis) and [strong](https://github.com/syntax-tree/mdast#strong). I think we should enforce this here.

src/layouts/Titled/components/TOC.tsx
austin94: Yes! I actually already had a component for this, but had missed it. Now sharing logic between the home page, desktop sidebar, mobile sidebar, and TOC sections.

src/components/MarkdownProvider/index.tsx
jzempel: I can buy this. Thanks for the explanation.

src/components/MarkdownProvider/components/Code.tsx
jzempel: Right, forgot about that.

talia-eisenberg:
image

Is the logo centered with the text? I think it is and likely playing visual tricks, but want to make sure.

austin94:
> Is the logo centered with the text? I think it is and likely playing visual tricks, but want to make sure.

@talia-eisenberg it is centered with flex-box right now. I think there may be some off-centering in the SVG, but that would need to be addressed in `svg-icons`.

talia-eisenberg:
@austin94 sounds good, I will take a look at the svg and re-export if needed

PR #31, opened 2020-05-05
feat: add initial component doc content

rossmoody:
## Description

**The automated PR build is failing for me due to unauthorization: https://share.getcloudapp.com/nOu89EJW

The goal of this PR is to start populating the site with content from the Google docs in the smartest way possible to effectively tee up remaining Eng/Design work that is going to take more consideration and time to complete. The way I approached this was: 

1. Flow in the content from Avatar, Breadcrumb, and Anchor component Google docs
2. Keep the code example component shell but include a hyperlink to the static example of what is to go inside it
3. For static assets that will eventually go on the page, I am wiring them up to artboards in Abstract so when designers take more time to do polished graphics they will be imported automatically.

For sections that need more Eng attention, I am commenting with TODO. The TODO items for component pages are typically wiring up the:

- Configuration section
- API table
- Do/Don't component implementation
- Component Code examples and code
- When to use/When not to use structure

Was hoping for feedback on the approach taken here. If it's looking like a promising way to start contributing content to the site, I can continue helping in this way and soliciting help from designers for the static graphic portions. 
 
## Checklist

### DIFF
--- a/src/nav/components.yml
+++ b/src/nav/components.yml
@@ -4,6 +4,8 @@
       title: Overview
 - title: Components
   items:
+    - title: Breadcrumbs
+      id: /components/breadcrumbs
     - title: Buttons
       items:
         - id: /components/buttons/anchor
--- a/src/nav/components.yml
+++ b/src/nav/components.yml
@@ -6,5 +6,7 @@
   items:
     - title: Buttons
       items:
+        - id: /components/buttons/anchor
+          title: Anchor
         - id: /components/buttons/button
           title: Button
--- a/src/nav/components.yml
+++ b/src/nav/components.yml
@@ -4,6 +4,8 @@
       title: Overview
 - title: Components
   items:
+    - title: Avatar
+      id: /components/avatar
     - title: Breadcrumbs
       id: /components/breadcrumbs
     - title: Buttons
--- a/src/nav/components.yml
+++ b/src/nav/components.yml
@@ -6,5 +6,5 @@
   items:
     - title: Buttons
       items:
-        - id: /components/button
+        - id: /components/buttons/button
           title: Button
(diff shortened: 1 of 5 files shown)


austin94:
@jzempel are there permissions required for non-maintainers to get to the `deploy` step in CircleCI? I'm seeing it disabled in these runs.

src/nav/components.yml
```diff
@@ -4,7 +4,13 @@
...
+    - title: Avatar
+      id: /components/avatar
+    - title: Breadcrumbs
+      id: /components/breadcrumbs
     - title: Buttons
       items:
-        - id: /components/button
+        - id: /components/buttons/anchor
+          title: Anchor
+        - id: /components/buttons/button
```
austin94, line 15: Even though these segments are nested in the sidebar, do we want them to be under a `/components/buttons/*` url? It may make sense to have everything directly under `/components/*` if possible.

rossmoody: We don't want the URL structure to reflect the IA?

austin94: The sidebar is just a visual representation, we can have any IA that fits. With `components/buttons/*` that could imply that there is a root `components/buttons` page which includes an overview. Would we have `components/ranges/range` and `components/ranges/multi-thumb-range`? I'm not sure this scales to all packages.

i.e. `components/accordions/stepper`, `components/breadcrumbs`

rossmoody: Great, great callout. Will amend.

rossmoody:
Closing this PR and parting it out into individual branches/initiatives. Henry and I met and decided to leverage the initial Avatar MDX from this to finish initial shared component setup outlined in [this Jira ticket.](https://zendesk.atlassian.net/browse/GARDEN-1195)

rossmoody:
Gonna keep this branch active for just a bit. There is a possibility we might be able to leverage the other two MDX component files when the time comes.

PR #32, opened 2020-05-06
fix: deploy context and error reporting

jzempel:
## Description

Use `writer` env variable context for deploy and exit with error if deployment fails.


jzempel:
thanks @rossmoody for helping to prove that the new config works as expected.

PR #20, opened 2020-04-07
feat: add package and prop-sheet sourcing with code examples

austin94:
## Description

This PR adds:
- `react-components` submodule
  - The git submodule is pinned to our most recent release. As we upgrade the site examples we can also upgrade the submodule tag in the same PR. This will keep prop-sheets in-sync.
- `ComponentTemplate.tsx`
  - We now source mdx files in the `content/components` directory
  - This includes it's own `nav.yml` file and local example directories
- `gatsby-source-react-components`
  - With us programmatically creating mdx pages we are unable to use GraphQL queries in our pages. This has required us to gather package and prop sheet information in a slightly obtuse way.
  - The `package` value in front-matter will provide a `props = { package: GardenReactPackage}` value to be rendered in our mdx files.
  - The `propSheets` value in front-matter will provide `props = { propSheets: [GardenReactPropSheet]}`
- `CodeExample`
  - Syntax Highlighting
    - This is using a standard PrismJS theme for now. Will be updated in a future PR.
  - Fork in Codesandbox
    - Uses the Codesandbox API to auto-generate isolated sandboxes for consumers that want to fork examples.
 
## Out of Scope

The visual styling for the typography, code example wrapper, and the examples have not been defined. There will be follow-on PRs to address all styling related issues as they are defined.

## Checklist

- [ ] :ok_hand: website updates are Garden Designer approved (add the designer as a reviewer)
- [ ] ~:black_nib: copy updates are approved (add the content strategist as a reviewer)~
- [ ] :wheelchair: analyzed via [axe](https://www.deque.com/axe/) and evaluated using VoiceOver
- [ ] :zap: audited via [web.dev](https://web.dev/measure/) for performance and SEO
- [ ] :memo: tested in Chrome, Firefox, Safari, Edge, and IE11

PR #34, opened 2020-05-11
feat: add avatar component page

hzhu:
## Description

This pull request introduces the Avatar Component Page. This component page is complete with all the expected shared components and live component examples.

## Checklist

* IE11 wasn't loading the localhost:8000. Will try with Netlify deployed demo.


src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,168 @@
...
+  }
+`;
+
+export const BestPracticeSection: React.FC<{
+  type: 'do' | 'dont' | 'caution';
+  imageSource?: string;
+}> = props => {
+  const data = useStaticQuery(graphql`
+    {
+      allFile(filter: { name: { regex: "/components-avatar/" } }) {
```
austin94, line 113: This regex will limit the `BestPracticeSection` to only working with avatar abstract images. Since static queries are unable to use variables to filter against I suggest raising this to a page-level query and providing the `publicUrl` to the prop directly.

src/components/MarkdownProvider/components/Usage.tsx
```diff
@@ -0,0 +1,17 @@
...
+import React from 'react';
+import styled from 'styled-components';
+
+const StyledUsage = styled.div`
+  @media (min-width: ${p => p.theme.breakpoints.sm}) {
+    display: flex;
+  }
+`;
+
+export const Usage: React.FC = props => <StyledUsage {...props} />;
```
austin94, line 17: Is it possible to export `StyledUsage` directly?

src/components/MarkdownProvider/index.tsx
```diff
@@ -59,12 +61,19 @@ export const MarkdownProvider: React.FC = ({ children }) => (
     <GlobalStyle />
     <MDXProvider
       components={{
+        /**
+         * Garden components
+         */
+        MD,
```
austin94, line 67: We shouldn't be providing `MD` components directly. Even in JSX/HTML code we should use `p, 2-h6, etc.` which gets mapped to the corresponding styling given by this provider. 

This is especially important to avoid `<div>` and `<span>` usage. Since this is text content we need the semantic meaning of the elements.

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,107 @@
...
+  Avatars display typography, icons, or images in different sizes. They help users quickly recognize
+  an entity.
+package: '@zendeskgarden/react-avatars'
+propSheets:
+  - avatars/src/elements/Avatar.tsx
+---
+
+<Usage>
+  <div>
+    <MD isBold>When to use</MD>
```
austin94, line 13: Use the `<strong>` element provided by the MarkdownProvider.

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,107 @@
...
+    <ul type="disc">
```
austin94, line 14: This should be consistent with our other `<ul>` usages in markdown. If `type="disc"` should be the default we can default that attribute in `typography/index.ts`

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,107 @@
...
+    <ul>
+      <li>Use the rounded square shape to represent objects, brands, or products</li>
+    </ul>
+  </BestPracticeSection>
+</BestPractice>
+
+## Package
+
+<PackageDescription data={props.data.mdx.package} />
+<PropSheets data={props.data.mdx.propsSheets} reactPackage={props.data.mdx.package} />
```
austin94, line 99: Should we have an `h2` saying `API` or `Prop Sheets` so the prop sheets are more identifiable from the side-bar?

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,168 @@
...
+    }
+  }
+
+  @media (max-width: ${p => p.theme.breakpoints.sm}) {
+    margin-bottom: ${props => props.theme.space.base * 4}px;
+  }
+`;
+
+export const BestPracticeSection: React.FC<{
+  type: 'do' | 'dont' | 'caution';
```
austin94, line 108: Rather than a single `BestPracticeSection` component. Would it be useful to have a `<Do>`, `<Dont>`, and `<Caution>` component to allow this to be more composable?

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+const StyledBestPracticeContainer = styled.div`
+  display: flex;
+  flex-basis: 0;
+  flex-direction: column;
+  flex-grow: 1;
+  margin-bottom: ${props => props.theme.space.base * 12}px;
+
+  @media (min-width: ${p => p.theme.breakpoints.sm}) {
+    margin: ${props => props.theme.space.base * 4}px ${props => props.theme.space.base * 2}px;
```
austin94, line 91: Are there other `theme.space.*` values that match these? We use sm/md/lg throughout the other templates.

src/components/MarkdownProvider/index.tsx
hzhu: This should be a `<b>` and not `MD`.

src/pages/components/avatar.mdx
austin94: Actually, if this is consistent between all usages, this can be an `MD isBold`, but it would be great if it could be abstracted into the component in some way along with the "When not to use"

src/components/MarkdownProvider/components/BestPractice.tsx
jzempel: Reminder: https://github.com/zendeskgarden/website/pull/20#discussion_r408383576

src/examples/avatar/AvatarStatus.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <img
+              alt=""
+              src="https://garden.zendesk.com/react-components/avatars/images/avatar-3.png"
+            />
+          </Avatar>
+        </StyledCol>
+        <StyledCol>
+          <Avatar backgroundColor={iconColor} size="large" status="away">
+            <img
+              alt=""
```
jzempel, line 44: Please provide meaningful alt text – "standard avatar", "available avatar" etc. – in this case it would most likely prove helpful.

src/examples/avatar/AvatarType.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const Example = () => {
+  const iconColor = DEFAULT_THEME.palette.grey[600];
+
+  return (
+    <Grid>
+      <Row alignItems="center">
+        <StyledCol>
+          <Avatar backgroundColor={iconColor} size="large">
+            <UserStrokeIcon role="img" aria-label="User" />
```
jzempel, line 28: `aria-label="icon avatar"`

src/examples/avatar/AvatarType.tsx
```diff
@@ -0,0 +1,49 @@
...
+          </Avatar>
+        </StyledCol>
+        <StyledCol>
+          <Avatar backgroundColor={iconColor} size="large">
+            <img
+              alt="avatar"
```
jzempel, line 34: `alt="image avatar"`

src/pages/components/avatar.mdx
hzhu: I used `Button` component page as a guide so didn't include "API" or "Prop Sheets". This is a good idea though, and I will make sure to update this accordingly in both component pages.

src/examples/avatar/AvatarShape.tsx
```diff
@@ -0,0 +1,40 @@
...
+import React from 'react';
+import styled from 'styled-components';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { Avatar } from '@zendeskgarden/react-avatars';
+import { Grid, Row, Col } from '@zendeskgarden/react-grid';
+import { ReactComponent as ZendeskIcon } from '@zendeskgarden/svg-icons/src/26/zendesk.svg';
+import { ReactComponent as UserStrokeIcon } from '@zendeskgarden/svg-icons/src/16/user-solo-stroke.svg';
+
+const StyledCol = styled(Col)`
+  margin-bottom: ${p => p.theme.space.sm};
```
jzempel, line 17: Why is a margin being added to the bottom?

src/pages/components/avatar.mdx
austin94: @rossmoody do you think this would be worthwhile for all the component pages?

src/components/MarkdownProvider/components/Usage.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export interface IUsageSectionProps {
+  type: string;
+}
+
+export const UsageSection: React.FC<{
+  type: string;
```
jzempel, line 25: Can you come up with a `boolean` prop name? Ideally for the negative rendering. i.e. `isNotRecommended`. Defer to @rossmoody for naming help :wink:

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,177 @@
...
+  li:not(:last-child) {
+    border-bottom: ${props =>
+      `${props.theme.borders.sm} ${getColor('neutralHue', 300, props.theme)}`};
+  }
+`;
+
+const StyledIcon = styled(({ children, ...props }) =>
+  cloneElement(React.Children.only(children), props)
+)`
+  margin-right: ${props => props.theme.space.xs};
```
jzempel, line 56: Since it's fairly trivial, please follow best practice and apply RTL logic to these `website` custom components.

src/examples/avatar/AvatarShape.tsx
jzempel: If `margin` is needed for multiple rows, then it should be applied to the top of all `<Row>` components except for the first child. But in this case (and several that follow) there is only one row.

src/examples/avatar/AvatarShape.tsx
```diff
@@ -0,0 +1,40 @@
...
+  text-align: center;
```
jzempel, line 18: Since there is such an appetite for this, we should probably add a `textAlign` prop to the grid `<Col>` component. With this and the comment above, a lot of the `StyledCol` usage in the website becomes redundant.

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,105 @@
...
+      <li>To display truncated text a native `title` attribute should be used</li>
+    </ul>
+  </UsageSection>
+</Usage>
+
+## Examples
+
+### Shape
+
+An Avatar can be a circle or rounded square. View the shape usage best practices.
```
jzempel, line 30: Link to "best practices"?

src/examples/avatar/AvatarType.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Avatar } from '@zendeskgarden/react-avatars';
+import { Grid, Row, Col } from '@zendeskgarden/react-grid';
+import { ReactComponent as UserStrokeIcon } from '@zendeskgarden/svg-icons/src/16/user-solo-stroke.svg';
+
+const StyledCol = styled(Col)`
+  margin-bottom: ${p => p.theme.space.sm};
+  text-align: center;
+`;
+
+const Example = () => {
```
jzempel, line 20: Rework this example to follow the text content and show "away, available, or active" in that order.

src/examples/avatar/AvatarType.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const iconColor = DEFAULT_THEME.palette.grey[600];
+
+  return (
+    <Grid>
+      <Row alignItems="center">
+        <StyledCol>
+          <Avatar backgroundColor={iconColor} size="large">
```
jzempel, line 27: Should these be the default size (remove `size` prop)?

src/examples/avatar/AvatarStatus.tsx
```diff
@@ -0,0 +1,54 @@
...
+`;
+
+const Example = () => {
+  const iconColor = DEFAULT_THEME.palette.grey[600];
+
+  return (
+    <Grid>
+      <Row alignItems="center">
+        <StyledCol>
+          <Avatar backgroundColor={iconColor} size="large">
```
jzempel, line 26: Should these be the default size (remove `size` prop)?

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,105 @@
...
+  <BestPracticeSection type="do" imageSource="components-avatar-shape-circle">
+    <ul>
+      <li>Use a circular shape to represent a person.</li>
+    </ul>
+  </BestPracticeSection>
+  <BestPracticeSection type="dont" imageSource="components-avatar-shape-square">
+    <ul>
+      <li>Use the rounded square shape to represent objects, brands, or products</li>
+    </ul>
+  </BestPracticeSection>
```
jzempel, line 91: This is intended to be a `do` section.

src/components/MarkdownProvider/components/Usage.tsx
jzempel: Actually, the preferred treatment would be to follow @austin94's [suggestion](https://github.com/zendeskgarden/website/pull/34#discussion_r423270284) on `BestPractice` and add first-class `<Use>` and `<Misuse>` components.

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,177 @@
...
+      }
+    }
+  `);
+
+const staticQueries: Record<string, Function> = {
+  avatar: useAvatarFiles
+};
+
+export const BestPracticeSection: React.FC<{
+  type: 'do' | 'dont' | 'caution';
```
jzempel, line 122: Don't want to lose @austin94 [suggestion](https://github.com/zendeskgarden/website/pull/34#discussion_r423270284) of adding `<Do>`, `<Dont>` and `<Caution>` components. That makes for a nicer API.

src/pages/components/avatar.mdx
austin94: For links that don't exist yet, we can have a local hyperlink `[Foobar](#)`. We will just need to clean these up before a wider release.

src/examples/avatar/AvatarShape.tsx
austin94: These styles are really dependent on content. I'm not sure we can accurately say how all content could be handled in RTL for example.  Should `text-align: left` always go to `right` in RTL? And vice-versa for other styles?

src/components/MarkdownProvider/components/BestPractice.tsx
austin94: > it's fairly trivial

There are many elements that don't handle RTL right now. With no other pages supporting it I don't think we should start introducing new standards until we know that RTL is a global feature that we will need.

src/pages/components/avatar.mdx
hzhu: Paired with @rossmoody on this, and we think so! I'm going to leave this off (for now). Ross is working on updating the heading hierarchy on a separate PR. Once Ross has outlined the heading structure, I'll update this 💯 .

src/components/MarkdownProvider/components/BestPractice.tsx
hzhu: Page-level queries ftw 🎖️, thanks!

src/examples/avatar/AvatarShape.tsx
hzhu: I spoke with @rossmoody about the `margin-bottom`. I think he said it's not needed, so I removed it. Well have to remove it from the button examples too 👍.

src/pages/components/avatar.mdx
hzhu: Updated with a placeholder

src/examples/avatar/AvatarType.tsx
hzhu: Yes I think so. I went ahead and removed the `size` prop.

src/examples/avatar/AvatarType.tsx
hzhu: ESLint is saying the word "image" would be redundant.

Screen Shot 2020-05-12 at 11 06 45 AM

src/components/MarkdownProvider/components/Usage.tsx
hzhu: Could you clarify what is meant by this? I initially assumed you meant that I can export `StyledUsage` directly like this:

```jsx
export const Usage = styled.div`
  @media (min-width: ${p => p.theme.breakpoints.sm}) {
    display: flex;
  }
`;
```

Is this close to what you had in mind?

src/components/MarkdownProvider/components/Usage.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const Misuse: React.FC = props => (
+  <div>
+    <MD isBold>When NOT to use</MD>
+    {props.children}
+  </div>
+);
+
+export const Usage = styled.div`
+  @media (min-width: ${p => p.theme.breakpoints.sm}) {
+    display: flex;
```
rossmoody, line 28: @jzempel This is more personal curiosity. When I was tinkering with this, I threw some Grid usage against the wall and I was curious if we are open to that? I don't think it's blocking but I'd love to [get some space in between the columns so they don't butt up against eachother. ](https://share.getcloudapp.com/o0ugOp25)

I had something like this to [create space between and below](https://share.getcloudapp.com/WnuGXdxK): 
```
  display: grid;
  grid-gap: ${p => p.theme.space.base * 6}px;
  grid-template-columns: repeat(auto-fill, minmax(300px, 1fr));
  margin: 0 0 ${p => p.theme.space.base * 9}px;
```
We can ofcourse do this with Flex (and prob should if this is a no-go).

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,119 @@
...
+    </ul>
+  </Do>
+  <Dont imageSource={props.data.avatarShapeCircle.children[0].publicURL}>
+    <ul>
+      <li>Use the rounded square shape to represent objects, brands, or products</li>
+    </ul>
+  </Dont>
+</BestPractice>
+
+## Package
```
rossmoody, line 94: This can be: 

```
## Configuration

<PackageDescription data={props.data.mdx.package} />

### API

<PropSheets data={props.data.mdx.propsSheets} reactPackage={props.data.mdx.package} />
```

src/components/MarkdownProvider/components/Usage.tsx
hzhu: @rossmoody, what do you think about using Garden's `react-grid` package to achieve the layout? It would also address the spacing issue you've discovered. I see two benefits of this: 1) Full support for IE11 2) More Garden usage. Let me know what you think.

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  li:not(:last-child) {
+    border-bottom: ${props =>
+      `${props.theme.borders.sm} ${getColor('neutralHue', 300, props.theme)}`};
+  }
+`;
+
+const StyledBestPracticeImageContainer = styled.div`
+  display: flex;
+  justify-content: center;
```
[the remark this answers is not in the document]
hzhu: Thanks, Safari fixed! 

Are there other issues with the images? They appear fine now (after `align-items: center;`). I left off `object-fit` since it's not supported in IE11, but I'll definitely address outstanding issues with the images.

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,163 @@
...
+  cloneElement(React.Children.only(children), props)
+)`
+  margin-right: ${props => props.theme.space.xs};
+  color: ${props => getColor(props.hue, 500, props.theme)};
+`;
+
+const StyledTitle = styled.div`
+  display: flex;
+  align-items: center;
+  margin-bottom: ${props => props.theme.space.xs};
```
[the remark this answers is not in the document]
hzhu: Updated, thanks 🙏 !

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,163 @@
...
+    display: block;
+  }
+`;
+
+const StyledBestPracticeContainer = styled.div`
+  display: flex;
+  flex-basis: 0;
+  flex-direction: column;
+  flex-grow: 1;
+  margin-bottom: ${props => props.theme.space.xxl};
```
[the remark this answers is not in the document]
hzhu: Updated, thanks!

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,122 @@
...
+    <ul>
+      <li>Use a circular shape to represent a person.</li>
+    </ul>
+  </Do>
+  <Dont imageSource={props.data.avatarShapeCircle.children[0].publicURL}>
+    <ul>
+      <li>Use the rounded square shape to represent objects, brands, or products</li>
+    </ul>
+  </Dont>
+</BestPractice>
```
hzhu, line 92: @jzempel @austin94 

There were some discussions of dot notation of these components:

```jsx
<BestPractice>
  <BestPractice.Do>
     // ...
  </BestPractice.Do>
  <BestPractice.Dont>
     // ...
  </BestPractice.Dont>
</BestPractice>
```

However, I ran into some issues exporting and reading the dot notation sub-component in the MDX files. I'll continue to look into this though.

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,122 @@
...
+Circular Avatars are used to represent a person because the shape is most often associated with a
+profile picture. Rounded squares are used for system applications like objects, brands, or products.
+This helps the user quickly distinguish between a person or a system when scanning.
+
+<BestPractice>
+  <Do imageSource={props.data.avatarShapeCircle.children[0].publicURL}>
+    <ul>
+      <li>Use a circular shape to represent a person.</li>
+    </ul>
+  </Do>
```
hzhu, line 86: @jzempel @austin94 

There were some discussions of dot notation of these components:

```jsx
<BestPractice>
  <BestPractice.Do>
     // ...
  </BestPractice.Do>
  <BestPractice.Dont>
     // ...
  </BestPractice.Dont>
</BestPractice>
```

However, I ran into some issues exporting and reading the dot notation sub-component in the MDX files. I'll continue to look into this though.

src/components/MarkdownProvider/components/Usage.tsx
austin94: Yep! I would suggest that we keep the `Styled*` naming to help know if we're able to reference them within other styled components.

We can rename it to `Usage` within the `MarkdownProvider` components prop.

src/examples/avatar/AvatarType.tsx
austin94: Maybe something like `Example avatar`? The `img` tag will read as "image" already

src/examples/avatar/AvatarShape.tsx
austin94: With the `Grid` updates being released my comment is now out-of-date 😄

src/pages/components/avatar.mdx
austin94: Interesting. If you're unable to get it working I think blaming MDX is totally valid and we can move away from the dot notation 😄

src/examples/avatar/AvatarType.tsx
hzhu: It has been updated to "example avatar". Let me know if there's another suggestion 👍.

src/examples/avatar/AvatarShape.tsx
hzhu: A new commit updates the `react-grid` package and leverages the `textAlign` prop on `Col`.

src/examples/avatar/AvatarType.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <Row alignItems="center">
+        <Col textAlign="center">
+          <Avatar backgroundColor={iconColor}>
+            <UserStrokeIcon role="img" aria-label="icon avatar" />
+          </Avatar>
+        </Col>
+        <Col textAlign="center">
+          <Avatar backgroundColor={iconColor}>
+            <img
+              alt="example avatar"
```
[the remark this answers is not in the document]
hzhu: Since it's not helpful for this particular context, I added a local ESLint ignore.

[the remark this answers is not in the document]
jzempel: Let's not pollute our examples with eslint ignore comments. Please add to the repo's `.eslintrc.json`.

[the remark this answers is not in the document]
jzempel: Every "alt" & "aria-label" in this repo will be reviewed and very intentional. If we decide to use "image", then it's for a good reason. While the rule is helpful, it is not useful in this context.

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+Circular Avatars are used to represent a person because the shape is most often associated with a
+profile picture. Rounded squares are used for system applications like objects, brands, or products.
+This helps the user quickly distinguish between a person or a system when scanning.
+
+<BestPractice>
+  <Do imageSource={props.data.avatarShapeCircle.children[0].publicURL}>
+    <p>Use a circular shape to represent a person.</p>
+  </Do>
+  <Dont imageSource={props.data.avatarShapeCircle.children[0].publicURL}>
```
hzhu, line 85: This should source from the square version of the image. The image source needs to be changed to `avatarShapeSquare`.

src/components/MarkdownProvider/components/Usage.tsx
rossmoody: Great solve 👍

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,165 @@
...
+const Section: React.FC<ISectionProps> = props => (
+  <StyledBestPracticeContainer>
+    <StyledBestPracticeImageContainer>
+      {props.imageSource && <img src={props.imageSource} alt={props.altText} />}
+    </StyledBestPracticeImageContainer>
+    <StyledWell hue={props.hue}>
+      <StyledTitle>
+        <StyledIcon hue={props.hue}>{props.icon}</StyledIcon>
+        <MD isBold>{props.title}</MD>
+      </StyledTitle>
```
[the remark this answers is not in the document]
hzhu: @jzempel I used the code from [jzempel/theme-bestpractice](https://github.com/zendeskgarden/website/tree/jzempel/theme-bestpractice) branch to align with what you had in mind. I made some minor adjustments:

* Use the `isRecessed` prop on `Well`.
* Factor `BestPractice` to handle cases with no backing `imageSource`.

I pushed these new commits up. Let me know what you think.

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+<BestPractice>
+  <Do imageSource={props.data.avatarShapeCircle.children[0].publicURL}>
+    <ul>
+      <li>Use a circular shape to represent a person.</li>
+    </ul>
+  </Do>
+  <Dont imageSource={props.data.avatarShapeCircle.children[0].publicURL}>
+    <ul>
+      <li>Use the rounded square shape to represent objects, brands, or products</li>
```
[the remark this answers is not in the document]
hzhu: > The words don't jive with the image.

There was a mistake in the image URL. Instead of `avatarShapeCircle`, it should have been `avatarShapeSquare`. Now that I've fixed the image URL, do the words match the image?

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,122 @@
...
+  <Use>
+    <ul>
+      <li>To briefly explain the function of a user interface element</li>
+      <li>To give contextual help or information regarding ambiguous actions</li>
+    </ul>
+  </Use>
+  <Misuse>
+    <ul>
+      <li>To display actions vital to using the UI</li>
+      <li>To display truncated text a native `title` attribute should be used</li>
```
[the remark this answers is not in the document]
hzhu: It looks like I mistakenly added the Tooltip usages. I've fixed this to have the Avatar usage text.

src/examples/avatar/AvatarStatus.tsx
```diff
@@ -0,0 +1,48 @@
...
+          </Avatar>
+        </Col>
+        <Col textAlign="center">
+          <Avatar backgroundColor={iconColor} status="available">
+            <img
+              alt="available avatar"
+              src="https://garden.zendesk.com/react-components/avatars/images/avatar-3.png"
+            />
+          </Avatar>
+        </Col>
```
jzempel, line 42: Treatment still not right. Should be away (yellow), available (green), active (red badge). Update alt text to match.

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,111 @@
...
+profile picture. Rounded squares are used for system applications like objects, brands, or products.
+This helps the user quickly distinguish between a person or a system when scanning.
+
+<BestPractice>
+  <Do imageSource={props.data.avatarShapeCircle.children[0].publicURL}>
+    Use a circular shape to represent a person.
+  </Do>
+  <Dont imageSource={props.data.avatarShapeSquare.children[0].publicURL}>
+    Use the rounded square shape to represent objects, brands, or products.
+  </Dont>
```
jzempel, line 80: This is still _very_ confusing. The words don't match the icon don't match the image. I'm pretty sure this is supposed to be a `<Do>`.

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,140 @@
...
+  }
+`;
+
+export const Section: React.FC<ISectionProps> = props => {
+  const content = (
+    <>
+      <Title>
+        <StyledIcon hue={props.hue}>{props.icon}</StyledIcon>
+        {props.title}
+      </Title>
```
jzempel, line 97: `<Title>` needs styled amendment to use flex display so that it can vertically center the icon.

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,140 @@
...
+const StyledWell = styled(props => <Well isRecessed {...props} />).attrs(p => ({
+  forwardedAs: p.tag
+}))<IStyledWellProps>`
+  border: none;
+  border-top: ${p => p.theme.borders.md};
+  border-radius: 0;
+  border-color: ${p => getColor(p.hue, 500, p.theme)};
+  padding: ${p => p.theme.space.md};
+
+  ${p => p.tag !== 'figcaption' && WellContainerStyles}
```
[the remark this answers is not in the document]
hzhu: Since the HTML spec said the `figure` can be "used to annotate illustrations, diagrams, photos, code listings, etc." I decided to not render a `figure` when an image didn't exist. Is it okay to render a `figure` / `figcaption` when an image doesn't exist? If that's okay I can consolidate the styles, and have a single `StyledFigure`.

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -0,0 +1,140 @@
...
+    title="Caution"
+    altText="caution"
+    hue="warningHue"
+    icon={<AlertErrorStrokeIcon />}
+    {...props}
+  />
+);
+
+export const BestPractice: React.FC = props => (
+  <StyledBestPractice>{props.children}</StyledBestPractice>
```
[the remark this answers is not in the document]
hzhu: I've swapped this out with a `Grid` implementation in a new commit. Let me know if this is close to what you had in mind.

src/components/MarkdownProvider/components/BestPractice.tsx
hzhu: Pushed a new commit which aligns the icon and `Title`.

hzhu:
> Still curious about https://github.com/zendeskgarden/website/pull/34#pullrequestreview-412170081. What's going on?

It looks like the git submodule may need to be updated to a newer commit hash. I'm investigating this now.

src/examples/avatar/AvatarStatus.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <Col textAlign="center">
+          <Avatar backgroundColor={iconColor} status={'active' as any} badge={8}>
```
jzempel, line 36: Please remove the `status` prop here. We shouldn't be showing non-existent prop values.

src/pages/components/avatar.mdx
```diff
@@ -0,0 +1,111 @@
...
+This helps the user quickly distinguish between a person or a system when scanning.
+
+<BestPractice>
+  <Do imageSource={props.data.avatarShapeCircle.children[0].publicURL}>
+    Use a circular shape to represent a person.
+  </Do>
+  <Do imageSource={props.data.avatarShapeSquare.children[0].publicURL}>
+    Use the rounded square shape to represent objects, brands, or products.
+  </Do>
+</BestPractice>
```
jzempel, line 81: After having seen the design options for this, I think the `<BestPractice>` component is going to need rework. But I don't want to continue to get in the way of this PR. Let's figure it out and plan for later.

src/components/MarkdownProvider/components/CodeExample/utils/retrieveCodesandboxParameters.ts
```diff
@@ -61,7 +61,7 @@ const packageJson = `
     "@zendeskgarden/react-buttons": "^8.0.0",
     "@zendeskgarden/react-chrome": "^8.0.0",
     "@zendeskgarden/react-dropdowns": "^8.0.0",
-    "@zendeskgarden/react-grid": "^8.0.0",
+    "@zendeskgarden/react-grid": "latest",
```
jzempel, line 64: Are we sure this is something we want to do? I'm concerned about selecting a semver that supports the mental mapping for developers looking at these examples in-depth. @austin94 thoughts?

austin94: It seems that Codesandbox doesn't map to the latest valid semver range when `^8.0.0` is used. `latest` is common in other repos that use this functionality so it may be our only options.

austin94: Maybe we should move all Garden dependencies in this to use `latest`?

jzempel: Yeah, at this point I would vote for consistency. Although, we're going to have to keep this in mind when moving to a new major version because the old version examples will tend to break.

hzhu: Moving all of them to the latest was my initial inclination. I can move them all to latest, and keep an eye out for major releases of `react-components` 👍.

jzempel: Another option would be to bump all the minors. `^8.9.0` Then this won't automatically break back-versions when we publish a v9.0 of the website.

jzempel: btw, I don't think this has been called out. But whatever is happening here is definitely a hackaround for a Codesandbox bug. Here's the latest dependency list with `^8.0.0`:

Screen Shot 2020-05-20 at 2 06 21 PM

And then with `latest`:

Screen Shot 2020-05-20 at 2 07 59 PM

Both are showing the use of version `8.10.0`. But only the `latest` supports recent grid props, etc.

jzempel: Seems to be a cache fault with CodeSandbox. When I change the grid range to `^8.x` everything seems to work. For now, I'd recommend setting all of these to `^8.x`.

Screen Shot 2020-05-20 at 2 18 57 PM

hzhu: Thanks, @jzempel! I'll update the react-component deps to ^8.x.

PR #49, opened 2020-05-27
feat: add anchor component page

rossmoody:
## Description

This pull request introduces the Anchor Component Page. Use the `View deployment` button further down this page to view a build of this Pull Request.

## Checklist

### DIFF
--- a/src/examples/anchor/AnchorDanger.tsx
+++ b/src/examples/anchor/AnchorDanger.tsx
@@ -9,7 +9,11 @@ import React from 'react';
 import { Anchor } from '@zendeskgarden/react-buttons';
 
 const Example = () => {
-  return <Anchor isDanger>Go to the Garden</Anchor>;
+  return (
+    <Anchor isDanger href="#">
+      Go to the Garden
+    </Anchor>
+  );
 };
 
 export default Example;
--- a/src/examples/anchor/AnchorDanger.tsx
+++ b/src/examples/anchor/AnchorDanger.tsx
@@ -1,0 +1,15 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Anchor } from '@zendeskgarden/react-buttons';
+
+const Example = () => {
+  return <Anchor isDanger>Go to the Garden</Anchor>;
+};
+
+export default Example;
--- a/src/examples/anchor/AnchorDanger.tsx
+++ b/src/examples/anchor/AnchorDanger.tsx
@@ -10,8 +10,8 @@ import { Anchor } from '@zendeskgarden/react-buttons';
 
 const Example = () => {
   return (
-    <Anchor isDanger href="#">
-      Go to the Garden
+    <Anchor isDanger href="#danger">
+      Leave the Garden
     </Anchor>
   );
 };
--- a/src/examples/anchor/AnchorDefault.tsx
+++ b/src/examples/anchor/AnchorDefault.tsx
@@ -9,7 +9,7 @@ import React from 'react';
 import { Anchor } from '@zendeskgarden/react-buttons';
 
 const Example = () => {
-  return <Anchor>Go to the Garden</Anchor>;
+  return <Anchor href="#">Go to the Garden</Anchor>;
 };
 
 export default Example;
--- a/src/examples/anchor/AnchorDefault.tsx
+++ b/src/examples/anchor/AnchorDefault.tsx
@@ -1,0 +1,15 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Anchor } from '@zendeskgarden/react-buttons';
+
+const Example = () => {
+  return <Anchor>Go to the Garden</Anchor>;
+};
+
+export default Example;
--- a/src/examples/anchor/AnchorDefault.tsx
+++ b/src/examples/anchor/AnchorDefault.tsx
@@ -9,7 +9,7 @@ import React from 'react';
 import { Anchor } from '@zendeskgarden/react-buttons';
 
 const Example = () => {
-  return <Anchor href="#">Go to the Garden</Anchor>;
+  return <Anchor href="#default">Go to the Garden</Anchor>;
 };
 
 export default Example;
--- a/src/examples/anchor/AnchorExternal.tsx
+++ b/src/examples/anchor/AnchorExternal.tsx
@@ -1,0 +1,15 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Anchor } from '@zendeskgarden/react-buttons';
+
+const Example = () => {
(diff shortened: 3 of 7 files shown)


src/examples/anchor/AnchorDanger.tsx
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Anchor } from '@zendeskgarden/react-buttons';
+
+const Example = () => {
+  return <Anchor isDanger>Go to the Garden</Anchor>;
```
rossmoody, line 12: Is it helpful to provide an href declaration in the examples?

```
<Anchor isDanger href="#">
  Go to the Garden
</Anchor>
```

austin94: Without the `href` the element won't be within the tab order. I would include it for all examples.

src/pages/components/anchor.mdx
```diff
@@ -0,0 +1,74 @@
...
+propSheets:
+  - buttons/src/elements/Anchor.tsx
+---
+
+import AnchorDefault from '../../examples/anchor/AnchorDefault.tsx';
+import AnchorDefaultCode from '!!raw-loader!../../examples/anchor/AnchorDefault.tsx';
+import AnchorExternal from '../../examples/anchor/AnchorExternal.tsx';
+import AnchorExternalCode from '!!raw-loader!../../examples/anchor/AnchorExternal.tsx';
+import AnchorDanger from '../../examples/anchor/AnchorDanger.tsx';
+import AnchorDangerCode from '!!raw-loader!../../examples/anchor/AnchorDanger.tsx';
```
austin94, line 15: Could we include these imports next to their respective examples? With the `raw-loader` usage this helps with co-location.

src/pages/components/anchor.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+An Anchor is used as a link to signal navigation.
+
+<CodeExample code={AnchorDefaultCode}>
+  <AnchorDefault />
+</CodeExample>
+
+### External Anchor
+
+Shows an icon at the end of the link and opens it in a new window or tab.
```
austin94, line 46: The `isExternal` prop doesn't actually open the link in a new window or tab. That's dependent on the `target` attribute. This wording may need to be tweaked.

src/pages/components/anchor.mdx
rossmoody: I will of course defer, but are we open to this approach in any way? Eventually (probably sooner than we think) we will have contributors on these pages that aren't Devs and so every effort we can make to have the Markdown/content feel like Markdown without import statement or other parts to ignore would be amazing from a contributor standpoint.

src/pages/components/anchor.mdx
rossmoody: Good callout. Is there a reason why we don't make that functionality part of the prop?

src/pages/components/anchor.mdx
hzhu: My preference is for co-location too since it helps me keep track of the example and their respective imports. If I need to make adjustments to either an import or its example, I would have to use mental energy to recall where each is in the file. It would require lower cognition to have them colocated. 

I think contributors need to be comfortable with `.mdx` files to contribute code changes. In these cases, they should be comfortable copying/pasting/moving the import statements. If they are only updating copy text, then it shouldn't be a concern. I think in the future when we define contributors (only those with Abstract tokens?) we could make adjustments.

src/pages/components/anchor.mdx
hzhu: > Is there a reason why we don't make that functionality part of the prop?

Good question 🤔. I think it's because `target` is a native `<a>` attribute. So no need for a prop, one can simply pass the `target` attribute directly. However, showing an icon is not native behavior, so we use a prop to customize and allow for rendering an icon.

src/pages/components/anchor.mdx
```diff
@@ -0,0 +1,74 @@
...
+    <ul>
+      <li>To navigate from page to page</li>
+      <li>To display links inline with text</li>
+      <li>To navigate within a page</li>
+    </ul>
+  </Use>
+  <Misuse>
+    <ul>
+      <li>
+        To prompt an action or submit data, use a <a href="/components/button">Button</a>
```
hzhu, line 28: It looks like `<a>` has yet to be mapped to markdown yet. Would you like to try mapping it so that it renders a Garden `<Anchor>`?

The existing mappings can be found [here](https://github.com/zendeskgarden/website/blob/master/src/components/MarkdownProvider/index.tsx#L78).

src/pages/components/anchor.mdx
austin94: It's difficult to infer that a specific `target` value is always "external" as well. Depending on where the link is shown (like in an iframe) it's not something that we can reliably enforce.

src/pages/components/anchor.mdx
rossmoody: I think I'm solving for problems we don't have again. I'm sure over time we'll uncover optimizations that simplify this syntax anyway. Will amend

src/pages/components/anchor.mdx
rossmoody: It is so great we're uncovering these things. These are improvements to the code and communication that I think will really help adoption.

src/pages/components/anchor.mdx
rossmoody: Serendipitous that we're doing the Anchor page and needed this. Felt only right that I take a stab at it. Please lemme know how I fared.

src/examples/anchor/AnchorDanger.tsx
```diff
@@ -0,0 +1,19 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Anchor } from '@zendeskgarden/react-buttons';
+
+const Example = () => {
+  return (
+    <Anchor isDanger href="#">
```
jzempel, line 13: @rossmoody I think it would be nifty if these examples linked to their respective sections. So, in this case `href="#danger"` – would help to prevent context jolt. Also, is there an opportunity to use copy other than "Go to the Garden"? (which doesn't sound very dangerous :wink:).

src/examples/anchor/AnchorDefault.tsx
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Anchor } from '@zendeskgarden/react-buttons';
+
+const Example = () => {
+  return <Anchor href="#">Go to the Garden</Anchor>;
```
jzempel, line 12: `href="#default"`

src/examples/anchor/AnchorExternal.tsx
```diff
@@ -0,0 +1,19 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Anchor } from '@zendeskgarden/react-buttons';
+
+const Example = () => {
+  return (
+    <Anchor isExternal href="#" target="_blank">
```
jzempel, line 13: Maybe make this `href="https://garden.zendesk.com/"`?

src/examples/anchor/AnchorDanger.tsx
rossmoody: Love both points. @vsri1137 I'll take a stab, but if you weigh in here, we can update.

PR #52, opened 2020-06-01
feat: add select component page

hzhu:
## Description

This PR introduces the Select component page.

## Checklist

### DIFF
--- a/src/examples/dropdowns/SelectDefault.tsx
+++ b/src/examples/dropdowns/SelectDefault.tsx
@@ -6,7 +6,6 @@
  */
 
 import React, { useState } from 'react';
-import styled from 'styled-components';
 import { Dropdown, Field, Menu, Item, Select, Label } from '@zendeskgarden/react-dropdowns';
 
 interface IItem {
@@ -20,13 +19,6 @@ const items = [
   { label: 'Succulent', value: 'item-3' }
 ];
 
-const StyledField = styled(Field)`
-  * {
-    /* stylelint-disable-next-line declaration-no-important */
-    appearance: none !important;
-  }
-`;
-
 const Example = () => {
   const [selectedItem, setSelectedItem] = useState(items[0]);
 
@@ -36,10 +28,10 @@ const Example = () => {
       onSelect={setSelectedItem}
       downshiftProps={{ itemToString: (item: IItem) => item && item.label }}
     >
-      <StyledField>
+      <Field>
         <Label>Plants</Label>
         <Select>{selectedItem.label}</Select>
-      </StyledField>
+      </Field>
       <Menu>
         {items.map(option => (
           <Item key={option.value} value={option}>
--- a/src/examples/dropdowns/SelectDefault.tsx
+++ b/src/examples/dropdowns/SelectDefault.tsx
@@ -1,0 +1,60 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { Dropdown, Field, Menu, Item, Select, Label } from '@zendeskgarden/react-dropdowns';
+
+interface IItem {
+  label: string;
+  value: string;
+}
+
+const items = [
+  { label: 'Item 1', value: 'item-1' },
+  { label: 'Item 2', value: 'item-2' },
+  { label: 'Item 3', value: 'item-3' }
+];
+
+const StyledSelect = styled(Select)`
+  * {
+    box-sizing: initial;
+  }
+`;
+
+const StyledField = styled(Field)`
+  * {
+    /* stylelint-disable-next-line declaration-no-important */
+    appearance: none !important;
+  }
+`;
+
+const Example = () => {
+  const [selectedItem, setSelectedItem] = useState(items[0]);
+
+  return (
+    <Dropdown
+      selectedItem={selectedItem}
+      onSelect={setSelectedItem}
+      downshiftProps={{ itemToString: (item: IItem) => item && item.label }}
+    >
+      <StyledField>
+        <Label>Label</Label>
+        <StyledSelect>{selectedItem.label}</StyledSelect>
+      </StyledField>
+      <Menu>
+        {items.map(option => (
+          <Item key={option.value} value={option}>
+            {option.label}
+          </Item>
+        ))}
+      </Menu>
+    </Dropdown>
+  );
+};
+
+export default Example;
--- a/src/examples/dropdowns/SelectDefault.tsx
+++ b/src/examples/dropdowns/SelectDefault.tsx
@@ -20,12 +20,6 @@ const items = [
   { label: 'Item 3', value: 'item-3' }
 ];
 
-const StyledSelect = styled(Select)`
-  * {
-    box-sizing: initial;
-  }
-`;
-
 const StyledField = styled(Field)`
   * {
     /* stylelint-disable-next-line declaration-no-important */
@@ -44,7 +38,7 @@ const Example = () => {
     >
       <StyledField>
         <Label>Label</Label>
-        <StyledSelect>{selectedItem.label}</StyledSelect>
+        <Select>{selectedItem.label}</Select>
       </StyledField>
       <Menu>
         {items.map(option => (
--- a/src/examples/dropdowns/SelectDefault.tsx
+++ b/src/examples/dropdowns/SelectDefault.tsx
@@ -15,9 +15,9 @@ interface IItem {
 }
 
 const items = [
-  { label: 'Item 1', value: 'item-1' },
-  { label: 'Item 2', value: 'item-2' },
-  { label: 'Item 3', value: 'item-3' }
+  { label: 'Cactus', value: 'item-1' },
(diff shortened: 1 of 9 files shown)


src/examples/dropdowns/SelectValidation.tsx
```diff
@@ -0,0 +1,79 @@
...
+  Menu,
+  Item,
+  Select,
+  Label,
+  Message
+} from '@zendeskgarden/react-dropdowns';
+
+const StyledSelect = styled(Select)`
+  * {
+    box-sizing: initial;
```
hzhu, line 23: The CSS for the select element makes the dropdown chevron icon very small (almost can't see it) when bedrock sets this element's `box-sizing` to `border-box`:

Screen Shot 2020-06-01 at 10 21 38 AM

When the `box-sizing` is reset to initial, the chevrons become the correct size:

Screen Shot 2020-06-01 at 10 22 04 AM

Any suggestions for an alternative approach here?

src/examples/dropdowns/SelectValidation.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+`;
+
+const StyledField = styled(Field)`
+  * {
+    /* stylelint-disable-next-line declaration-no-important */
+    appearance: none !important;
```
hzhu, line 30: It seems like some CSS from bedrock is causing the styles to render this in Safari:

Screen Shot 2020-06-01 at 10 13 32 AM

Targeting the faux input with `appearance: none !important;` results in the following:

Screen Shot 2020-06-01 at 10 14 41 AM

Having a `StyledField` with a `!important` to target an element is probably not ideal. Any suggestions here?

src/examples/dropdowns/SelectValidation.tsx
jzempel: This is fixed by https://github.com/zendeskgarden/react-components/pull/756

hzhu: Awesome, I've removed the `StyledSelect`. I expect to update the `react-dropdowns` package once https://github.com/zendeskgarden/react-components/pull/756 is released. That should fix the chevron sizing issue.

src/pages/components/select.mdx
```diff
@@ -0,0 +1,101 @@
...
+import SelectSize from '../../examples/dropdowns/SelectSize.tsx';
+import SelectSizeCode from '!!raw-loader!../../examples/dropdowns/SelectSize.tsx';
+
+<CodeExample code={SelectSizeCode}>
+  <SelectSize />
+</CodeExample>
+
+### Disabled
+
+Mark the input disabled to restrict user interaction. [Read more](#best-practices) about best
```
hzhu, line 80: The best practice section for this page is currently punted.  @rossmoody do we know if that will be included in the initial release?

src/pages/components/select.mdx
```diff
@@ -0,0 +1,101 @@
...
+  </Use>
+  <Misuse>
+    <ul>
+      <li>
+        To filter a large list of options, consider an{' '}
+        <a href="/components/autocomplete">Autocomplete</a>
+      </li>
+      <li>
+        To make multiple selection from a list, use a{' '}
+        <a href="/components/multiselect">Multiselect</a>
```
hzhu, line 25: The Autocomplete page is being designed and development should begin shortly. The Multiselect component page is under development. I can change the href to `href="#TODO"` if preferable.

src/examples/dropdowns/SelectSize.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+const StyledField = styled(Field)`
+  * {
+    /* stylelint-disable-next-line declaration-no-important */
+    appearance: none !important;
+  }
+`;
+
+const Example = () => {
+  const [selectedItem, setSelectedItem] = useState(items[0]);
```
hzhu, line 32: This is a single state that powers both dropdowns. Feels a bit strange, so I'm going to have state for each of the dropdown sizes.

src/examples/dropdowns/SelectDefault.tsx
```diff
@@ -0,0 +1,54 @@
...
+  { label: 'Orchid', value: 'item-2' },
+  { label: 'Succulent', value: 'item-3' }
+];
+
+const StyledField = styled(Field)`
+  * {
+    /* stylelint-disable-next-line declaration-no-important */
+    appearance: none !important;
+  }
+`;
```
jzempel, line 28: Let's remove from here and fix in `react-components`.

src/examples/dropdowns/SelectDisabled.tsx
```diff
@@ -0,0 +1,30 @@
...
+import React from 'react';
+import styled from 'styled-components';
+import { Dropdown, Field, Select, Label } from '@zendeskgarden/react-dropdowns';
+
+const StyledField = styled(Field)`
+  * {
+    /* stylelint-disable-next-line declaration-no-important */
+    appearance: none !important;
+  }
+`;
```
jzempel, line 17: ditto

src/examples/dropdowns/SelectHint.tsx
```diff
@@ -0,0 +1,55 @@
...
+  { label: 'Orchid', value: 'item-2' },
+  { label: 'Succulent', value: 'item-3' }
+];
+
+const StyledField = styled(Field)`
+  * {
+    /* stylelint-disable-next-line declaration-no-important */
+    appearance: none !important;
+  }
+`;
```
jzempel, line 28: ditto

src/examples/dropdowns/SelectSize.tsx
```diff
@@ -0,0 +1,81 @@
...
+  { label: 'Orchid', value: 'item-2' },
+  { label: 'Succulent', value: 'item-3' }
+];
+
+const StyledField = styled(Field)`
+  * {
+    /* stylelint-disable-next-line declaration-no-important */
+    appearance: none !important;
+  }
+`;
```
jzempel, line 29: ditto

src/examples/dropdowns/SelectValidation.tsx
jzempel: Let's remove all the `StyledField` overrides and address this in `react-components` :wink:

src/pages/components/select.mdx
jzempel: I think this is the right option as it prevents future clean-up.

src/examples/dropdowns/SelectSize.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <Menu isCompact>
+              {items.map(option => (
+                <Item key={option.value} value={option}>
+                  {option.label}
+                </Item>
+              ))}
+            </Menu>
+          </Dropdown>
+        </Col>
+        <Col>
```
jzempel, line 57: Recommend flipping column order so that the example order corresponds with the wording order: "Select can be default or compact in size."

rossmoody:
> 3\. A comment/question for @rossmoody and @vsri1137. I think another "When not to use" would be when there are only 2 options to choose between, in which case, a Radio should be used instead. But perhaps this is a best practice?

Henry brought up a pretty good point where our Menus are used for actions and selections and  since this rule only applies to when the `Menu` is serving actions and not options, we moved it to punt to be in a best practice where we can elaborate a little more.

src/pages/components/select.mdx
```diff
@@ -0,0 +1,101 @@
...
+    </ul>
+  </Use>
+  <Misuse>
+    <ul>
+      <li>
+        To filter a large list of options, consider an{' '}
+        <a href="/components/autocomplete">Autocomplete</a>
+      </li>
+      <li>
+        To make multiple selection from a list, use a{' '}
```
allisonacs, line 24: Should be "To make multiple selections" (selection needs to be made plural)

allisonacs:
> @hzhu I was confused and thought we were on the Menu page. That sentiment _does_ fit appropriately on this Component page, right? Because in this application the Menu will only ever be a `listbox`.

@rossmoody @hzhu sorry, I'm confused :/ What is the decision here?

rossmoody:
We originally had that "Best practice" in the `Menu` documentation, but Henry pointed out it's not entirely appropriate there because the Menu can contain actions or selectable options depending on the trigger. So in some Menus we wouldn't say use to use a Radio cause the items might be actions. 

But in this case, in the `Select` component page, I think it totally works cause the Menu is only ever going to contain options which can translate well to Radios. But since there is some possible elaboration and consideration on the right place this guidance should go, could we punt it for now?

hzhu:
@rossmoody @allisonacs 

`Select` component usages with `Menu` can be used for _selection_. However, `Trigger` usages with `Menu` is not used for `Selection`. Since we are talking about the `Select` component in this PR, we could put the recommendation to use a radio group in certain contexts under "When to use" _or  "Best practices". Personally, I think it should be under best practices as a Select dropdown with only two components is _valid_, but not recommended.

>  I think another "When not to use" would be when there are only 2 options to choose between, in which case, a Radio should be used instead. But perhaps this is a best practice?

Yeah this is totally fair. I'll leave it up to ya'll to decide which section this goes it.

src/pages/components/select.mdx
hzhu: Copy has been updated, thanks!

allisonacs:
OK, cool, sounds like the consensus is to punt on this / keep it in a forthcoming best practices session. So let's just go with that. No further action needed.

PR #53, opened 2020-06-01
feat(components): add configuration section

rossmoody:
## Description

This PR introduces the configuration section styling and logic into the component pages. 

## Detail

1. The `UnorderedList.Item` component renders a wrapping child div. Instead of creating an additional wrapping element for Flex styling, we styled the child div with Flex styles using `& > div` CSS selector
1. The Configuration and API section props and component names were getting a little confusing. The names were revised a bit to [more accurately reflect what they are for.](https://github.com/zendeskgarden/website/compare/rmoody/configuration?expand=1#diff-0b9a85d38fa919055fd1752418521e3aR85)

## Checklist

- [x] :zap: audited via [web.dev](http://localhost:8000/components/button//) for performance and SEO

### DIFF
--- a/src/components/MarkdownProvider/components/Configuration.tsx
+++ b/src/components/MarkdownProvider/components/Configuration.tsx
@@ -21,17 +21,17 @@ export interface IPackage {
 
 const StyledUnorderedList = styled(UnorderedList)`
   margin: 0 0 ${p => p.theme.space.xxl};
-  border-bottom: solid 1px ${p => getColor('grey', 300, p.theme)};
+  border-bottom: ${p => `${p.theme.borders.sm} ${getColor('grey', 300, p.theme)}`};
 `;
 
 const StyledListItem = styled(UnorderedList.Item)`
   list-style: none;
 
-  & > :first-child {
+  & > div {
     display: flex;
-    margin: ${p => p.theme.space.base * 3}px 0;
-    border-top: solid 1px ${p => getColor('grey', 300, p.theme)};
-    padding: ${p => p.theme.space.base * 3}px 0 0;
+    margin: ${p => p.theme.space.sm} 0;
+    border-top: ${p => `${p.theme.borders.sm} ${getColor('grey', 300, p.theme)}`};
+    padding: ${p => p.theme.space.sm} 0 0;
     font-family: ${p => p.theme.fonts.mono};
   }
 `;
@@ -43,7 +43,7 @@ const StyledListItemLabel = styled.label`
 `;
 
 const StyledSpan = styled(Span)`
-  margin-right: ${p => p.theme.space.base * 3}px;
+  margin-right: ${p => p.theme.space.sm};
 `;
 
 export const Configuration: React.FC<{ reactPackage: IPackage; propSheets: ComponentDoc[] }> = ({
@@ -54,7 +54,6 @@ export const Configuration: React.FC<{ reactPackage: IPackage; propSheets: Compo
     <StyledUnorderedList>
       <StyledListItem>
         <StyledListItemLabel>Name</StyledListItemLabel>
-        {/* This is showing the version on this page and not the latest available. Need to fix */}
         <StyledSpan>{reactPackage.version}</StyledSpan>
         <StyledSpan>•</StyledSpan>
         <StyledSpan>
--- a/src/components/MarkdownProvider/components/Configuration.tsx
+++ b/src/components/MarkdownProvider/components/Configuration.tsx
@@ -34,10 +34,8 @@ const StyledListItem = styled(UnorderedList.Item)`
   }
 `;
 
-const StyledListItemLabel = styled.label`
+const StyledListItemLabel = styled(MD)`
   min-width: ${p => p.theme.space.base * 20}px;
-  font-family: ${p => p.theme.fonts.system};
-  font-weight: ${p => p.theme.fontWeights.semibold};
 `;
 
 const StyledDot = styled(Span)`
@@ -56,7 +54,7 @@ export const Configuration: React.FC<{ reactPackage: IPackage; propSheets: Compo
   return (
     <StyledUnorderedList>
       <StyledListItem>
-        <StyledListItemLabel>Name</StyledListItemLabel>
+        <StyledListItemLabel isBold>Name</StyledListItemLabel>
         {reactPackage.version}
         <StyledDot>•</StyledDot>
         <Anchor
@@ -72,17 +70,17 @@ export const Configuration: React.FC<{ reactPackage: IPackage; propSheets: Compo
         </Anchor>
       </StyledListItem>
       <StyledListItem>
-        <StyledListItemLabel>Install</StyledListItemLabel>
+        <StyledListItemLabel isBold>Install</StyledListItemLabel>
         <StyledMD isMonospace>npm install {reactPackage.name}</StyledMD>
       </StyledListItem>
       <StyledListItem>
-        <StyledListItemLabel>Deps</StyledListItemLabel>
+        <StyledListItemLabel isBold>Deps</StyledListItemLabel>
         <StyledMD isMonospace>
           npm install react react-dom prop-types styled-components @zendeskgarden/react-theming
         </StyledMD>
       </StyledListItem>
       <StyledListItem>
-        <StyledListItemLabel>Import</StyledListItemLabel>
+        <StyledListItemLabel isBold>Import</StyledListItemLabel>
         <StyledMD isMonospace>
           import{' '}
           {`{ ${propSheets && propSheets.map(propSheet => propSheet.displayName).join(', ')} }`}{' '}
--- a/src/components/MarkdownProvider/components/Configuration.tsx
+++ b/src/components/MarkdownProvider/components/Configuration.tsx
@@ -21,11 +21,10 @@ interface IPackage {
 const StyledUnorderedList = styled(UnorderedList)`
   margin: 0 0 ${p => p.theme.space.xxl};
   border-bottom: ${p => `${p.theme.borders.sm} ${getColor('grey', 200, p.theme)}`};
+  list-style: none;
 `;
 
 const StyledListItem = styled(UnorderedList.Item)`
-  list-style: none;
-
   & > div {
     display: flex;
     margin: ${p => p.theme.space.base * 2.5}px 0;
@@ -43,7 +42,9 @@ const StyledDot = styled(Span)`
   color: ${p => getColor('grey', 600, p.theme)};
 `;
 
-const StyledMD = styled(MD)`
+const StyledMono = styled(MD)`
+  /* 1px nudge to align mono type with 14px label baseline */
+  line-height: 21px;
   color: ${p => getColor('grey', 700, p.theme)};
 `;
 
@@ -66,28 +67,28 @@ export const Configuration: React.FC<{ reactPackage: IPackage; propSheets: Compo
         <Anchor
           href={`https://www.npmjs.com/package/@zendeskgarden/react-${reactPackage.packageName}`}
         >
-          View on NPM
+          View on npm
         </Anchor>
       </StyledListItem>
       <StyledListItem>
         <StyledListItemLabel isBold>Install</StyledListItemLabel>
-        <StyledMD isMonospace>npm install {reactPackage.name}</StyledMD>
+        <StyledMono isMonospace>npm install {reactPackage.name}</StyledMono>
       </StyledListItem>
       <StyledListItem>
         <StyledListItemLabel isBold>Deps</StyledListItemLabel>
-        <StyledMD isMonospace>
+        <StyledMono isMonospace>
           npm install react react-dom prop-types styled-components @zendeskgarden/react-theming
-        </StyledMD>
+        </StyledMono>
       </StyledListItem>
       <StyledListItem>
         <StyledListItemLabel isBold>Import</StyledListItemLabel>
-        <StyledMD isMonospace>
+        <StyledMono isMonospace>
           import{' '}
           {`{ ${propSheets && propSheets.map(propSheet => propSheet.displayName).join(', ')} }`}{' '}
           from &apos;
           {reactPackage.name}
           &apos;
-        </StyledMD>
+        </StyledMono>
       </StyledListItem>
     </StyledUnorderedList>
   );
--- a/src/components/MarkdownProvider/components/Configuration.tsx
+++ b/src/components/MarkdownProvider/components/Configuration.tsx
@@ -7,7 +7,7 @@
 
 import React from 'react';
 import styled from 'styled-components';
-import { UnorderedList, Span } from '@zendeskgarden/react-typography';
+import { UnorderedList, Span, MD } from '@zendeskgarden/react-typography';
 import { Anchor } from '@zendeskgarden/react-buttons';
 import { getColor } from '@zendeskgarden/react-theming';
 import { ComponentDoc } from 'react-docgen-typescript';
@@ -31,7 +31,6 @@ const StyledListItem = styled(UnorderedList.Item)`
     margin: ${p => p.theme.space.base * 2.5}px 0;
     border-top: ${p => `${p.theme.borders.sm} ${getColor('grey', 200, p.theme)}`};
     padding: ${p => p.theme.space.base * 2.5}px 0 0;
-    font-family: ${p => p.theme.fonts.mono};
   }
 `;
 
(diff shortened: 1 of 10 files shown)


src/components/MarkdownProvider/components/Configuration.tsx
```diff
@@ -0,0 +1,95 @@
...
+        </StyledSpan>
+        <StyledSpan>•</StyledSpan>
+        <StyledSpan>
+          <Anchor
+            href={`https://www.npmjs.com/package/@zendeskgarden/react-${reactPackage.packageName}`}
+            target="_blank"
+          >
+            View on NPM
+          </Anchor>
+        </StyledSpan>
```
hzhu, line 75: Alternatively, we could have a single `<a>` tag without the wrapping `<span>`. We could remove all the `StyledSpan` elements and only style the two dots with horizontal margin of `12px`. It might look something [like this](https://5ed596ca6afa3c0a11941f33--upbeat-fermi-0d23ad.netlify.app/components/avatar#configuration).

src/components/MarkdownProvider/components/Configuration.tsx
```diff
@@ -0,0 +1,95 @@
...
+            View source
+          </Anchor>
+        </StyledSpan>
+        <StyledSpan>•</StyledSpan>
+        <StyledSpan>
+          <Anchor
+            href={`https://www.npmjs.com/package/@zendeskgarden/react-${reactPackage.packageName}`}
+            target="_blank"
+          >
+            View on NPM
```
hzhu, line 73: Should "View on NPM" and "View Source" be using mono typography even though these aren't code/commands that can be ran in the terminal?

src/components/MarkdownProvider/components/Configuration.tsx
```diff
@@ -0,0 +1,95 @@
...
+  return (
+    <StyledUnorderedList>
+      <StyledListItem>
+        <StyledListItemLabel>Name</StyledListItemLabel>
+        <StyledSpan>{reactPackage.version}</StyledSpan>
+        <StyledSpan>•</StyledSpan>
+        <StyledSpan>
+          <Anchor
+            href={`https://github.com/zendeskgarden/react-components/tree/master/packages/${reactPackage.packageName}`}
+            target="_blank"
```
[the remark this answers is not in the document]
austin94: If these were to open in a new tab we should be using the `isExternal` prop as well

src/components/MarkdownProvider/components/Configuration.tsx
```diff
@@ -0,0 +1,95 @@
...
+          </Anchor>
+        </StyledSpan>
+      </StyledListItem>
+      <StyledListItem>
+        <StyledListItemLabel>Install</StyledListItemLabel>
+        npm install {reactPackage.name}
+      </StyledListItem>
+      <StyledListItem>
+        <StyledListItemLabel>Deps</StyledListItemLabel>
+        npm install {reactPackage.peerDependencies.join(' ')}
```
austin94, line 83: We will always have the same peer-dependencies through the packages. Could we hard-code this as `styled-component` and `react-theming`?

src/components/MarkdownProvider/components/Configuration.tsx
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+import React from 'react';
+import styled from 'styled-components';
+import { UnorderedList, Span } from '@zendeskgarden/react-typography';
+import { Anchor } from '@zendeskgarden/react-buttons';
+import { getColor } from '@zendeskgarden/react-theming';
+import { ComponentDoc } from 'react-docgen-typescript';
+
+export interface IPackage {
```
austin94, line 15: Is this interface used in other areas? This shouldn't be exported

src/components/MarkdownProvider/components/Configuration.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <StyledListItemLabel>Deps</StyledListItemLabel>
+        npm install {reactPackage.peerDependencies.join(' ')}
+      </StyledListItem>
+      <StyledListItem>
+        <StyledListItemLabel>Import</StyledListItemLabel>
+        import{' '}
+        {`{ ${propSheets && propSheets.map(propSheet => propSheet.displayName).join(', ')} }`} from
+        &apos;
+        {reactPackage.name}
+        &apos;
```
austin94, line 91: These imports are normally shown at the `propSheet` level, not at the package configuration level. What is the intention with this line?

src/components/MarkdownProvider/components/PropSheets.tsx
```diff
@@ -9,25 +9,17 @@ import React from 'react';
...
 import { Table, Head, Body, HeaderRow, HeaderCell, Row, Cell } from '@zendeskgarden/react-tables';
-
-import { IPackage } from './PackageDescription';
 import { StyledH3 } from './Typography';
 
-export const PropSheets: React.FC<{ data: ComponentDoc[]; reactPackage: IPackage }> = ({
-  data,
-  reactPackage
-}) => {
+export const PropSheets: React.FC<{ propSheets: ComponentDoc[] }> = ({ propSheets }) => {
```
austin94, line 16: Could you go into depth on the prop rename for `data -> propSheets`? With the name of the component the usage will now be `<PropSheets propSheets={...} />` which feel repetitive

src/components/MarkdownProvider/components/Configuration.tsx
```diff
@@ -0,0 +1,95 @@
...
+    display: flex;
+    margin: ${p => p.theme.space.sm} 0;
+    border-top: ${p => `${p.theme.borders.sm} ${getColor('grey', 300, p.theme)}`};
+    padding: ${p => p.theme.space.sm} 0 0;
+    font-family: ${p => p.theme.fonts.mono};
+  }
+`;
+
+const StyledListItemLabel = styled.label`
+  min-width: ${p => p.theme.space.base * 20}px;
```
austin94, line 40: Should min-width values use `theme.breakpoint` widths?

src/components/MarkdownProvider/components/Configuration.tsx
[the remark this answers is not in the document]
rossmoody: @jzempel @allisonacs What do y'all think?

src/components/MarkdownProvider/components/Configuration.tsx
rossmoody: I think this is a call for @allisonacs in design definitions.

src/components/MarkdownProvider/components/Configuration.tsx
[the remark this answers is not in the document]
jzempel: Having external links pop out to `target="_blank"` makes sense to me, especially if we have consistent treatment throughout the website. But I'll defer to @allisonacs for the desired user experience. Sometimes loading up a bunch of tabs doesn't feel super nice. 🤷

src/components/MarkdownProvider/components/PropSheets.tsx
rossmoody: As Henry and I were working through, the naming of this section was confusing to understand so this hoped to address that a bit. The prop was `data` on `PackageDescription` but it's grabbing the package information. But then on `PropSheets` component the `data` property was grabbing something different and reactPackage is grabbing package information. Open to changes, but the hope was to make it more straightforward to read. 

```
<PackageDescription data={props.data.mdx.package} />
<PropSheets data={props.data.mdx.propsSheets} reactPackage={props.data.mdx.package} />
```

src/components/MarkdownProvider/components/Configuration.tsx
rossmoody: To show the import command next to the import label in the configuration section.

src/components/MarkdownProvider/components/Configuration.tsx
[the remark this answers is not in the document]
austin94: If we want to start using `target="_blank"` more actively I suggest we also add some security attributes as well.

https://medium.com/sedeo/how-to-fix-target-blank-a-security-and-performance-issue-in-web-pages-2118eba1ce2f

specifically the ` rel=”noopener noreferrer”` attribute

src/components/MarkdownProvider/components/Configuration.tsx
rossmoody: Would that look like?

```
npm install react react-dom prop-types styled-components @zendeskgarden/react-theming
```

austin94: That feels right to me. We could also include this once in the dev installation section that was planned at the top of the `Components` layouts. Not sure we need to repeat this on every page now that this info is all on a single site.

src/components/MarkdownProvider/components/Configuration.tsx
austin94: For some pages this will include 10+ components. This feels like duplicated info from the individual prop sheet area.

With us splitting packages between documentation pages this also only shows a subset of the available imports.

rossmoody: The import statements were pulled from above each individual API sheet and put in the Configuration section as a collection to see what's available as a group so there is one reference point.

[Here is the tentative design spec](https://share.getcloudapp.com/12uNeGeE) I was working against but @allisonacs may have additional revisions.

hzhu: @austin94 is your suggestion to remove this list item entirely? That makes sense to me, but I just wanted to confirm if I understood your comment.

src/components/MarkdownProvider/components/PropSheets.tsx
hzhu: I noticed that `propSheets` was passed as "data", but `package` is passed as "reactPackage". Why would one prop be named "data" but not the other? For example:

```tsx
<Configuration reactPackage={props.data.mdx.package} data={props.data.mdx.propsSheets} />
```

My suggestion was what Ross currently has in this PR, or:

```tsx
<Configuration data={...props.data.mdx.package, ...props.data.mdx.propsSheets} />
```

Let us know what you think @austin94. I'm down with reverting it to what you had originally 👍.

austin94: A single data prop makes sense to me. Feel free to go with whatever feels best.

src/components/MarkdownProvider/components/Configuration.tsx
austin94: That feels like a good path to me. Imports are used for including a specific component. I don't see anyone using this value directly.

rossmoody: I'm unsure what to do here. I think what we're proposing is a change from the design spec I was working with which is fine, but I think @allisonacs is taking that on. I will defer to direction on this. Please let me know if I should move the import statements back to the top of each API table component to match the current Garden website design or continue with the new design spec.

src/components/MarkdownProvider/components/Configuration.tsx
rossmoody: Much simpler imo. ty 👍

allisonacs:
> 1. Should we open the links in a [new window or same window?](https://github.com/zendeskgarden/website/pull/53#discussion_r433547749)

Same window. Users can always choose to open in a new window if they prefer. 

> 2. Should we change the formatting of the first list item [to be regular paragraph style?](https://github.com/zendeskgarden/website/pull/53#discussion_r433551341)

Yes, see updated design [on Abstract](https://share.goabstract.com/9df5eb02-b603-4604-9ae6-af6dbac4dc9a). Or here:

image

🚨 @rossmoody please note some slightly finicky bits:
- the mono font color is `grey-700`, not 800
- the mono font size is 13px, not 14px
- the bullet between the version number and the links is `grey-600`

(both the mono adjustments align it to the same type treatment used in our `code` component, just without the grey background)

> 3. Should we move the import statements to above each component or [keep them as a group in the configuration section?](https://github.com/zendeskgarden/website/pull/53#discussion_r433948830)

From an information organization standpoint, I think it is best to keep this as it is. However, once we release, we should talk with other developers and understand how (or if) this change impacts their workflow.

src/components/MarkdownProvider/components/Configuration.tsx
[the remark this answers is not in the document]
rossmoody: Opening in same window

src/components/MarkdownProvider/components/Configuration.tsx
rossmoody: Revised

rossmoody:
I think feedback has been addressed and this is ready for another look.

src/components/MarkdownProvider/components/PropSheets.tsx
jzempel, line 16: This component should probably be named `API`. Per below, it would make more sense in context. But that's effort for a separate PR.

```
## Configuration

<Configuration ... />

## API

<API propSheets={props.data.mdx.propsSheets} />
```

src/pages/components/button.mdx
```diff
@@ -61,14 +61,13 @@ import ButtonDangerCode from '!!raw-loader!../../examples/button/ButtonDanger.ts
...
 
-### Writing button labels
+<Configuration reactPackage={props.data.mdx.package} propSheets={props.data.mdx.propsSheets} />
 
-## Package
+## API
 
-<PackageDescription data={props.data.mdx.package} />
-<PropSheets data={props.data.mdx.propsSheets} reactPackage={props.data.mdx.package} />
+<PropSheets propSheets={props.data.mdx.propsSheets} />
```
jzempel, line 70: fwiw, doesn't have to happen in this PR. But we need to tee-up support for `@ignore` as handled by Styleguidist. https://react-styleguidist.js.org/docs/documenting/#ignoring-props. In this prop sheet, we shouldn't see `isSelected`. See https://github.com/zendeskgarden/react-components/blob/992c41ff83f4f6349bd491b6cdbe3dc9373b0fae/packages/buttons/src/elements/Button.tsx#L30-L31

If not already done, please capture this issue. It is a must-do for v1.

src/components/MarkdownProvider/components/Configuration.tsx
```diff
@@ -0,0 +1,94 @@
...
+const StyledListItem = styled(UnorderedList.Item)`
+  list-style: none;
+
+  & > div {
+    display: flex;
+    margin: ${p => p.theme.space.base * 2.5}px 0;
+    border-top: ${p => `${p.theme.borders.sm} ${getColor('grey', 200, p.theme)}`};
+    padding: ${p => p.theme.space.base * 2.5}px 0 0;
+  }
+`;
```
jzempel, line 35: Prefer using `UnorderedList type="none"` which probably means this component can go away and just provide a

```
const StyledListItemContent = styled.div`
  display: flex;
  ...
`
```

for the child content.

src/components/MarkdownProvider/components/Configuration.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <Anchor
+          href={`https://github.com/zendeskgarden/react-components/tree/master/packages/${reactPackage.packageName}`}
+        >
+          View source
+        </Anchor>
+        <StyledDot>•</StyledDot>
+        <Anchor
+          href={`https://www.npmjs.com/package/@zendeskgarden/react-${reactPackage.packageName}`}
+        >
+          View on NPM
```
jzempel, line 69: [~NPM~](https://www.npmjs.com/) refers to themselves as "npm"

src/pages/components/button.mdx
rossmoody: So we aren't bloating the MVP Story points up and down in Jira adding tasks, I have started collecting bugs, nits and issues like this in issue #1280. I know we will have to address it a different way, but I was capturing it there for now.

jzempel: Excellent. TY

src/components/MarkdownProvider/components/PropSheets.tsx
rossmoody: Capture in 1280

src/components/MarkdownProvider/components/Configuration.tsx
jzempel: Actually, if that suggestion :point_up: is a type error, you also add `list-style-type: none;` to `StyledUnorderedList`. Either way, you should be able to use `UnorderedList.Item` without wrapping.

src/components/MarkdownProvider/components/Configuration.tsx
rossmoody: Ty good spot

src/components/MarkdownProvider/components/Configuration.tsx
rossmoody: Moved the list style to the parent. To avoid div soup, kept the child div targeting.

jzempel: Thanks for the child `div` reminder, @rossmoody. Sounds good. Btw, that child div exists so that we can re-leverage baseline typography components (`SM`, `MD`, `LG`, etc) within the context of the list.

PR #56, opened 2020-06-04
feat: add tabs page

rossmoody:
## Description

This PR introduces the first iteration of the Tabs component page

## Detail

- I had to manually add the static assets in this case because there is currently an error pulling in Abstract assets. The [error I'm getting looks like this.](https://share.getcloudapp.com/v1ue6lzd)
- The Vertical Tabs example has an explicit `320px` wrapping div. This is to stop the example from shifting between each tab click. I'm unsure if this is the best solve, but I was rather enthused about keeping the component formatting out of the example code. 
- I commented in TODOs for Configuration and API section until the Configuration PR is merged. We can probably still make those updates in this PR. 

## Checklist


src/pages/components/tabs.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+### Vertical
+
+Tabs are horizontal by default but can be manually set to stack vertically.
+
+import TabsVertical from '../../examples/tabs/TabsVertical.tsx';
+import TabsVerticalCode from '!!raw-loader!../../examples/tabs/TabsVertical.tsx';
+
+<CodeExample code={TabsVerticalCode}>
+  <div style={{ width: '320px' }}>
```
austin94, line 46: Why do we have to provide a fixed-width for this example? If it's required lets include it within the `TabsVertical` implementation and use styled-component rather than inline styles.

src/pages/components/tabs.mdx
```diff
@@ -0,0 +1,103 @@
...
+</BestPractice>
+
+## Accessibility
+
+### Automatic activation
+
+Garden doesn't endorse the use of [automatic activation in
+tabs](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html). There is no way of
+guaranteeing that tab content will not require a network request or page refresh and it will
+inevitably cause inconsistency in tab interactions across products.
```
austin94, line 87: Does this need to be included? We enforce the other non-automatic interaction already.

 > is no way of guaranteeing that tab content will not require a network request or page refresh

The "page refresh" flow isn't a tab pattern. This can happen where consumers build custom tab visuals with anchors, but I don't believe this is applicable to Garden Tabs.

jzempel: We discussed this during review and thought it was helpful to include text indicating our stance on the matter so we could point if/when questions came up.

src/pages/components/tabs.mdx
```diff
@@ -0,0 +1,103 @@
...
+labels are internationalized and many languages contain more characters per word than English.
+
+Horizontal tabs do not wrap or overflow so anything that exceeds the width of the page will become
+truncated and may hide important actions.
+
+import tabHorizDo from '../../images/components-tabs-horizontal-do.png';
+import tabHorizDont from '../../images/components-tabs-horizontal-dont.png';
+
+<BestPractice>
+  <Do imageSource={tabHorizDo}>Use a circular shape to represent a person.</Do>
```
jzempel, line 76: Copy/paste

jzempel:
@rossmoody, in order to reference `props.data.avatarShapeCircle...` you need to add something like the following to your `tabs.mdx` page query:

```js
export const pageQuery = graphql`
  query($fileAbsolutePath: String) {
    ...SidebarPageFragment
    avatarShapeCircle: abstractAsset(layerName: { eq: "components-avatar-shape-circle" }) {
      children {
        ... on File {
          publicURL
        }
      }
    }
  }
`;
```

After that, a hard localhost restart is necessary as Gatsby's live server does not pick up the change.

src/pages/components/tabs.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+### Be mindful of horizontal tab length
+
+When more than 5 tabs are needed or the tab strings are lengthy, consider a vertical tab layout. Tab
+labels are internationalized and many languages contain more characters per word than English.
+
+Horizontal tabs do not wrap or overflow so anything that exceeds the width of the page will become
+truncated and may hide important actions.
+
+import tabHorizDo from '../../images/components-tabs-horizontal-do.png';
```
hzhu, line 72: > @hzhu let's figure out the Abstract issue before merging this one. @rossmoody how did you debug this in GraphiQL?

Yes, let's address this issue before merging. The Abstract API defaults the limit of assets to 50 so `components-tabs-horizontal-do` (along with many others) are not retrieved due to the limit.

We'll need to pass in a `limit` to the [abstractClient.assets.file](https://sdk.goabstract.com/docs/abstract-api/#list-assets-for-a-file) API call. It accepts a number. So an arbitrary high number, or [MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) can be used.

@austin94  @rossmoody

austin94: Let's follow the [Abstract pagination guidelines](https://sdk.goabstract.com/docs/pagination/). The default limit of 50 seems good to me to limit the length of calls.

We could just look loop through pages in our `abstract-asset` plugin.

src/pages/components/tabs.mdx
```diff
@@ -0,0 +1,103 @@
...
+      <li>As a secondary navigation bar that spans multiple pages</li>
+    </ul>
+  </Misuse>
+</Usage>
+
+## Examples
+
+### Default
+
+The basic usage of a Tab component.
```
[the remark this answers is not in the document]
rossmoody: Revised to `Tab layout is horizontal by default but can be set to stack vertically.` V, if you have a better please throw it in here and I'll update.

src/pages/components/tabs.mdx
rossmoody: It's kinda sounding like this sentiment, whether we keep it or not, is going to require group input, consideration and rewriting. I'll pull it out to punt section for now and then next rev we can work on it with V. (If this is a must-have, lemme know and we can make it happen)

src/pages/components/tabs.mdx
rossmoody: This was sloppy. It was a bandaid to stop this specific component from shifting when Tabs were clicked as the content changed. Good news is putting in more content into the TabPanel pushed the example full width and now the shift doesn't happen anymore. So this is gone.

src/pages/components/tabs.mdx
```diff
@@ -0,0 +1,103 @@
...
+<CodeExample code={TabsDisabledCode}>
+  <TabsDisabled />
+</CodeExample>
+
+## Best practices
+
+### Be mindful of horizontal tab length
+
+When more than 5 tabs are needed or the tab strings are lengthy, consider a vertical tab layout. Tab
+labels are internationalized and many languages contain more characters per word than English.
```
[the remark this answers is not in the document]
rossmoody: Updated for now pending future V revisions.

src/pages/components/tabs.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+Horizontal tabs do not wrap or overflow so anything that exceeds the width of the page will become
+truncated and may hide important actions.
+
+import tabHorizDo from '../../images/components-tabs-horizontal-do.png';
+import tabHorizDont from '../../images/components-tabs-horizontal-dont.png';
```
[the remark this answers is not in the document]
rossmoody: Super agree. Revised to be larger.

src/pages/components/tabs.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+<BestPractice>
+  <Do imageSource={tabHorizDo}>Use a circular shape to represent a person.</Do>
+  <Dont imageSource={tabHorizDont}>Use a long tab label that causes formatting problems.</Dont>
```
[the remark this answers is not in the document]
rossmoody: Revised for now pending V updates.

rossmoody:
Ok, aside from the Abstract automation functionality, I think the comments from this PR are addressed. @jzempel You think the Configuration PR will merge today? I could update that section here as well before merge.

src/pages/components/tabs.mdx
hzhu: @austin94 great idea! Thanks for sharing the docs. PR for fix is in https://github.com/zendeskgarden/website/pull/57.

rossmoody:
Ok we're all ready for another full look with Abstract automation and Configuration section. I noted the `@component` description that's showing up in the 1280 bug ticket.

.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ references:
   workspace_root: &workspace_root ~/project
 
   # Combination of https://circleci.com/orbs/registry/orb/circleci/node#commands-with-cache and https://www.gatsbyjs.org/docs/build-caching/#clearing-cache
-  gatsby_cache_key: &gatsby_cache_key gatsby-cache-{{ .Branch }}-{{ checksum "yarn.lock" }}-{{ checksum "gatsby-config.js" }}-{{ checksum "gatsby-node.js" }}
+  gatsby_cache_key: &gatsby_cache_key gatsby-cache-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "yarn.lock" }}-{{ checksum "gatsby-config.js" }}-{{ checksum "gatsby-node.js" }}
```
hzhu, line 14: @jzempel @austin94 

Since @rossmoody added static Abstract images in an earlier commit, the build ran into some caching issues that require us to clear the cache since those image files don't exist anymore. For example, see this [build log](https://app.circleci.com/pipelines/github/zendeskgarden/website/426/workflows/22f8f2b0-d482-473b-97aa-a0b0ca03316d/jobs/855).

We decided on adding a `CACHE_VERSION` number to bust the cache as outlined [here](https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-cache).

src/examples/tabs/TabsDisabled.tsx
```diff
@@ -0,0 +1,43 @@
...
+      </TabPanel>
+      <TabPanel item="tab-2">
+        The sugar maple is one of America’s most-loved trees. In fact, more states have claimed it
+        as their state tree than any other single species—for New York, West Virginia, Wisconsin,
+        and Vermont, the maple tree stands alone.
+      </TabPanel>
+      <TabPanel item="tab-3">
+        Cornus is a genus of about 30–60 species of woody plants in the family Cornaceae, commonly
+        known as dogwoods, which can generally be distinguished by their blossoms, berries, and
+        distinctive bark
```
hzhu, line 37: Should this paragraph end with a period like the other `Tab` paragraphs? 

> "...and distinctive bark."

src/examples/tabs/TabsVertical.tsx
```diff
@@ -0,0 +1,39 @@
...
+      </TabPanel>
+      <TabPanel item="tab-2">
+        The sugar maple is one of America’s most-loved trees. In fact, more states have claimed it
+        as their state tree than any other single species—for New York, West Virginia, Wisconsin,
+        and Vermont, the maple tree stands alone.
+      </TabPanel>
+      <TabPanel item="tab-3">
+        Cornus is a genus of about 30–60 species of woody plants in the family Cornaceae, commonly
+        known as dogwoods, which can generally be distinguished by their blossoms, berries, and
+        distinctive bark
```
hzhu, line 33: Should this paragraph end with a period like the other `Tab` paragraphs?

rossmoody: Good spot, ty

rossmoody:
Ok, all addressed. Logged the bug where the API section is rendering an empty table when there are no props in 1280 ticket.

jzempel:
@allisonacs fwiw, we've got a list of problems bringing the existing prop sheets in to the website. You can see how this is meant to render on the Styleguidist side of things. Generally, out-of-scope for this PR. But, in my mind, these are must-do's in order to get to a v1.

Screen Shot 2020-06-09 at 10 35 31 AM

rossmoody:
@allisonacs 
1. Amended pending V input. 
2-5. I pulled the best practice in the component out to the punt section for the sake of moving forward. 
6. Logged in ticket 1280
7. Logged in ticket 1280
8. Logged in 1280

PR #59, opened 2020-06-09
chore: information architecture links

rossmoody:
## Description
This PR aims to have every link on the site that is to be in the MVP represented and with a legitimate endpoint (whether on this site or the old). 

## Detail

1. I had to write a little logic to get the sidebar links to style and function correctly depending on whether they are an `<a>` or a Gatsby Link. Gatsby `<Link>` components only allow for internal linking. Very open to feedback on that implementation as it was a smidge outside my comfort zone (especially the types). 
2. On the homepage I omitted the "Site news" section as that won't be represented.
4. Pages that are for on the MVP site but aren't ready yet I created blank placeholder pages. 
5. I think the Content section is still set to be revised a little depending on @vsri1137 direction, but it's pretty close.
6. I created a Versions page to link to v6 and v7 versions of the old site in the Components section. This will need written but depending on how it's received here I will add that task to 1280. 

## Checklist

### DIFF
--- a/src/layouts/Root/components/Footer.tsx
+++ b/src/layouts/Root/components/Footer.tsx
@@ -6,15 +6,11 @@
  */
 
 import React from 'react';
-import styled, { css } from 'styled-components';
+import { css } from 'styled-components';
 import { math } from 'polished';
 import { getColor } from '@zendeskgarden/react-theming';
 import { ReactComponent as GardenIcon } from '@zendeskgarden/svg-icons/src/26/garden.svg';
 import MaxWidthLayout from 'layouts/MaxWidth';
-
-const StyledFooterItem = styled.div`
-  margin-right: ${p => p.theme.space.lg};
-`;
 
 const Footer: React.FC = () => (
   <footer
@@ -27,25 +23,6 @@ const Footer: React.FC = () => (
     `}
   >
     <MaxWidthLayout>
-      <div
-        css={css`
-          display: flex;
-          padding-bottom: ${p => p.theme.space.md};
-
-          @media (max-width: ${p => p.theme.breakpoints.md}) {
-            flex-direction: column;
-            padding-left: ${p => math(`${p.theme.iconSizes.md} + ${p.theme.space.md}`)};
-            text-align: center;
-          }
-        `}
-      >
-        <StyledFooterItem>Site Updates</StyledFooterItem>
-        <StyledFooterItem>License</StyledFooterItem>
-        <StyledFooterItem>Contribute</StyledFooterItem>
-        <StyledFooterItem>Blog</StyledFooterItem>
-        <StyledFooterItem>Github</StyledFooterItem>
-        <StyledFooterItem>v8.x.x</StyledFooterItem>
-      </div>
       <div
         css={css`
           display: flex;
--- a/src/layouts/Root/components/Footer.tsx
+++ b/src/layouts/Root/components/Footer.tsx
@@ -6,11 +6,18 @@
  */
 
 import React from 'react';
-import { css } from 'styled-components';
+import styled, { css } from 'styled-components';
 import { math } from 'polished';
 import { getColor } from '@zendeskgarden/react-theming';
 import { ReactComponent as GardenIcon } from '@zendeskgarden/svg-icons/src/26/garden.svg';
+import { Link } from './StyledNavigationLink'
 import MaxWidthLayout from 'layouts/MaxWidth';
+
+const StyledFooterItem = styled(Link)`
+  margin-right: ${ p => p.theme.space.lg};
+  color: ${ p => p.theme.colors.background};
+  cursor: pointer;
+`;
 
 const Footer: React.FC = () => (
   <footer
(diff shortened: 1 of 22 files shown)


austin94:
image

Is this the intended visualization? There seems to be some artifacts that need to be updated for all screen sizes. Would it be possible to remove the site news content, but still have a similar 2-column layout? This is a pretty big visual change.

>I omitted the footer links in this PR. I don't know if we ever really finalized that section but considering License, Site updates, Contribute (no great landing point), and Version pages don't really feel applicable at the moment down there I kept it plain for the time being.

Could this be handled in our bug bash? License, Github repo, and some other links are important for us to transition the domain over.

src/layouts/Root/components/StyledNavigationLink.tsx
```diff
@@ -5,10 +5,35 @@
...
 import { getColor } from '@zendeskgarden/react-theming';
 
+interface ILink {
+  children: string;
+  to: string;
+  activeClassName: string;
+}
+
+const Link = ({ children, to, ...props }: ILink) => {
+  const internal = /^\/(?!\/)/u.test(to);
```
austin94, line 20: There are some alternative strategies for determining relative vs absolute URLS. You may want to take a look at the logic in https://stackoverflow.com/questions/10687099/how-to-test-if-a-url-string-is-absolute-or-relative

src/pages/components/versions.mdx
```diff
@@ -0,0 +1,18 @@
...
+- [Versions 7](https://garden-v7.netlify.app/)
+- [Versions 6](https://garden-v6.netlify.app/)
+
+import { graphql } from 'gatsby';
+
+export const pageQuery = graphql`
+  query($fileAbsolutePath: String) {
+    ...SidebarPageFragment
+  }
+`;
```
jzempel, line 18: Feels like this could benefit from being under a `/pages/components/foundations/` subdirectory.

rossmoody:
> Could this be handled in our bug bash? License, Github repo, and some other links are important for us to transition the domain over.

I don't mind keeping links, I'm unsure what to keep and where it should go though. Would we be ok with omitting for now and then in the bug bash bringing back what we definitely want to go in there?

src/layouts/Root/components/StyledNavigationLink.tsx
rossmoody: Oh that's cool. Would we prefer that regex pattern?

src/layouts/Root/components/Footer.tsx
```diff
@@ -27,25 +23,6 @@ const Footer: React.FC = () => (
...
-            text-align: center;
-          }
-        `}
-      >
-        <StyledFooterItem>Site Updates</StyledFooterItem>
-        <StyledFooterItem>License</StyledFooterItem>
-        <StyledFooterItem>Contribute</StyledFooterItem>
-        <StyledFooterItem>Blog</StyledFooterItem>
-        <StyledFooterItem>Github</StyledFooterItem>
-        <StyledFooterItem>v8.x.x</StyledFooterItem>
```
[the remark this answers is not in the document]
rossmoody: Sounds good. @allisonacs Any way you can tell me what should stay and where it should go?

src/pages/components/versions.mdx
rossmoody: Visually it's [there in the sidebar](https://share.getcloudapp.com/d5uWYnQp), would we want to commit to that URL slug structure as well?

jzempel: I guess we're keeping the URL structure pretty flat in the other pillars. After looking closer, this should probably stay the way you have it.

src/layouts/Root/components/Footer.tsx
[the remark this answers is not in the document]
jzempel: Looping back, we can probably bag "License" from this list and link the "©️ Zendesk 2020" off to https://www.zendesk.com/legal/ just like zendesk.com does.

allisonacs:
Blog, Github, and Version should stay (assuming you have those links, I think you said you had links to v6 and 7 in a thread above).

allisonacs:
image

To get that column looking more considered, you can add back in the last link from our current site:
[Designing for inclusivity: How and why to get started](https://www.invisionapp.com/inside-design/designing-for-inclusivity/) by Allison Shaw

src/layouts/Root/components/Footer.tsx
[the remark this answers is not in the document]
allisonacs: I commented above but we should keep:
- Blog - goes to design.zendesk.com
- Github
- Versioning

I'll follow JZ's lead for licensing!

src/nav/components.yml
```diff
@@ -14,7 +24,75 @@
...
+        - id: https://zendeskgarden.github.io/react-components/grid
+          title: Dots
+        - id: https://zendeskgarden.github.io/react-components/grid
+          title: Progress
+        - id: https://zendeskgarden.github.io/react-components/grid
+          title: Skeleton
+        - id: https://zendeskgarden.github.io/react-components/grid
+          title: Spinner
+        - id: https://zendeskgarden.github.io/react-components/grid
+          title: Inline
```
allisonacs, line 72: These should be going to /loaders, not grid, right?

allisonacs:
We aren't launching with the Patterns section I thought?

src/nav/design.yml
```diff
@@ -16,3 +16,5 @@
           title: 26px
         - id: /design/icons/wordmarks
           title: Wordmarks
+    - title: Color
+      id: https://garden.zendesk.com/css-components/utilities/color/
```
jzempel, line 20: Although marginally less attractive, let's point this url to https://zendeskgarden.github.io/react-components/theming/#!/PALETTE. In contrast with the `css-components` page, the hex values displayed are computed directly from the theme palette and can be counted on for source accuracy.

src/layouts/Root/components/Footer.tsx
```diff
@@ -10,10 +10,13 @@ import styled, { css } from 'styled-components';
 import { math } from 'polished';
 import { getColor } from '@zendeskgarden/react-theming';
 import { ReactComponent as GardenIcon } from '@zendeskgarden/svg-icons/src/26/garden.svg';
+import { Link } from './StyledNavigationLink';
 import MaxWidthLayout from 'layouts/MaxWidth';
 
-const StyledFooterItem = styled.div`
+const StyledFooterItem = styled(Link)`
   margin-right: ${p => p.theme.space.lg};
+  cursor: pointer;
```
jzempel, line 18: Is `cursor: pointer` necessary here?

rossmoody: No! I wonder if I was battling with a cache issue for when these were still `div` elements. Great spot. tyty

rossmoody:
@allisonacs I agree, I don't like having to get to the bottom of the content. I didn't dive in but I think this might require possibly a larger refactor (it might not but my gut is telling me it might). Since this PR is already biting off a lot more than I originally intended to chew, could we add that fix to bug bash and let this one go?

allisonacs:
@rossmoody yea, right on. I pulled it into 1280. Will approve.

PR #62, opened 2020-06-10
feat: add text input component page

hzhu:
## Description

This PR introduces the Text Inputs component page.

## Checklist


src/examples/text-inputs/FauxInput.tsx
```diff
@@ -0,0 +1,18 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Field, Label, FauxInput } from '@zendeskgarden/react-forms';
+
+const Example = () => (
+  <Field>
+    <Label>Plant</Label>
+    <FauxInput />
```
austin94, line 14: Should this include content within the `FauxInput` to show the expected use?

src/examples/text-inputs/Bare.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Field, Label, Input } from '@zendeskgarden/react-forms';
+
+const Example = () => (
+  <Field>
+    <Label>Plant</Label>
+    <Input isBare placeholder="Plants" />
+  </Field>
+);
+
+export default Example;
```
hzhu, line 18: Note to self: changing `Bare.tsx` to `InputBare.tsx` would increase consistent naming across files. See avatar and button examples for consistent naming. This naming change would apply to all Text input example files.

src/nav/components.yml
```diff
@@ -22,3 +22,7 @@
           title: Multiselect
     - title: Tabs
       id: /components/tabs
+    - title: Text inputs
+      items:
+        - id: /components/text-inputs
+          title: Text inputs
```
jzempel, line 28: In conjunction with @austin94's "Input" comment, the IA feels off here. What is meant to be contained under this "Text inputs" parent?

src/examples/text-inputs/Size.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Grid, Row, Col } from '@zendeskgarden/react-grid';
+import { Field, Label, Hint, Input } from '@zendeskgarden/react-forms';
+
+const Example = () => (
+  <Grid>
+    <Row>
+      <Col>
+        <Field>
+          <Label>Plant</Label>
+          <Hint>Plants in the Garden</Hint>
```
jzempel, line 18: Recommend dropping the `Hint` here in order to make the example easier to compare.

src/pages/components/text-inputs.mdx
```diff
@@ -0,0 +1,199 @@
...
+
+<CodeExample code={SizeCode}>
+  <Size />
+</CodeExample>
+
+### Bare
+
+Input borders and padding can be removed using the `isBare` property. Use an
+[aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute)
+to label the input.
```
jzempel, line 115: > Use an aria-label

Why? Also not reinforced by the example.

src/examples/text-inputs/Inline.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Field, Label, Input } from '@zendeskgarden/react-forms';
+import styled from 'styled-components';
+
+const StyledField = styled(Field)`
+  display: flex;
+  align-items: center;
+`;
+
+const StyledLabel = styled(Label)`
+  /* stylelint-disable-next-line property-no-unknown */
```
jzempel, line 18: Please knock out the `property-no-unknown` rule in `website`'s stylelint config and remove the `stylelint-disable` comment here. Note that eslint/stylelint disable comments are distracting in example codez.

src/pages/components/text-inputs.mdx
```diff
@@ -0,0 +1,199 @@
...
+import Inline from '../../examples/text-inputs/Inline.tsx';
+import InlineCode from '!!raw-loader!../../examples/text-inputs/Inline.tsx';
+
+<CodeExample code={InlineCode}>
+  <Inline />
+</CodeExample>
+
+### Disabled
+
+A disabled Input is not interactive, and not perceived by a screen reader. See [Disabled UI](#TODO)
```
jzempel, line 175: What is the v1 plan for missing pattern guidelines?

src/pages/components/text-inputs.mdx
```diff
@@ -0,0 +1,199 @@
+---
+title: Text inputs
+description: >
+  A text input is a form field that lets the user enter text.
+package: '@zendeskgarden/react-forms'
+propSheets:
+  - forms/src/elements/Input.tsx
+  - forms/src/elements/Textarea.tsx
+  - forms/src/elements/MediaInput.tsx
+  - forms/src/elements/FauxInput.tsx
```
jzempel, line 10: Missing `Field`, `Label`, `Hint`, `Message`.

src/pages/components/text-inputs.mdx
```diff
@@ -0,0 +1,199 @@
...
+import ValidationCode from '!!raw-loader!../../examples/text-inputs/Validation.tsx';
+
+<CodeExample code={ValidationCode}>
+  <Validation />
+</CodeExample>
+
+### Inline
+
+An Input uses `display:block` by default. It can use `display:inline`. See example code for inline
+styling overrides.
```
jzempel, line 164: This sentence isn't accurate.

- a `Field` is `display: block` by default
- an `Input` is `display: inline-block` by default
- the associated demo is showing how a label can be inlined with an input via `display: flex`

src/nav/components.yml
hzhu: Oops, IA is off and I can make the appropriate adjustments. The concerns around "Text inputs" were brought up in the Google doc yesterday.

src/examples/text-inputs/Size.tsx
hzhu: This suggestion conflicts with the Google doc. Should we confirm with @ginnywood first?

src/pages/components/text-inputs.mdx
```diff
@@ -0,0 +1,199 @@
...
+</CodeExample>
+
+### Bare
+
+Input borders and padding can be removed using the `isBare` property. Use an
+[aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute)
+to label the input.
+
+import Bare from '../../examples/text-inputs/Bare.tsx';
+import BareCode from '!!raw-loader!../../examples/text-inputs/Bare.tsx';
```
allisonacs, line 118: Rather than repeating the form label in the placeholder text, let's use something that more explicit denotes that there is a text input there. Something like, "Type plant name"

src/pages/components/text-inputs.mdx
```diff
@@ -0,0 +1,199 @@
...
+
+<CodeExample code={BareCode}>
+  <Bare />
+</CodeExample>
+
+### Focus inset
+
+Box-shadow focus styling can be inset using the `focusInset` property.
+
+This keeps the focus contained within the input area when it is flush with another component.
```
allisonacs, line 128: Doesn't feel like this should be two separate lines. They are related thoughts, so no line break. cc @vsri1137 

Also the example feels like it's missing the described other component flush with the text input.

src/pages/components/text-inputs.mdx
hzhu: This concern was brought up yesterday in the Google doc. How does the following sound?

> An Input can be styled to have an inline layout. See example code for inline styling overrides.

If that copy is acceptable, then I can update this mdx file and make adjustments to the Google doc accordingly.

src/pages/components/text-inputs.mdx
hzhu: > Why? Also not reinforced by the example.

I brought this up in the Google doc and I am hoping to get the correct copy soon. Then the PR can align with the Google doc. I think this sentence about `aria-label` will be deleted.

src/pages/components/text-inputs.mdx
```diff
@@ -0,0 +1,199 @@
...
+<CodeExample code={PlaceholderCode}>
+  <Placeholder />
+</CodeExample>
+
+### Validation
+
+Success, warning, or danger validation messages are shown using the Message component with the text
+Input.
+
+import Validation from '../../examples/text-inputs/Validation.tsx';
```
allisonacs, line 154: image

The responsive resizing on this example feels off. Could we do something more like what we see on menus/selects? 

image

(should probably increase the vertical margin on those examples to give each field more breathing room as well.)

src/nav/components.yml
rossmoody: You can follow this IA structure for the placement of this: https://5ee276003c482c6da838de83--zendeskgarden.netlify.app/components

src/examples/text-inputs/Size.tsx
ginnywood: This was only a mistake we had a comment to correct in the final image assets.

src/pages/components/text-inputs.mdx
ginnywood: We talked about this in the doc review; I was showing a `bare` input with a label, and the team let me know that `bare` inputs typically have a hidden label and an aria-label. I left a comment to fix the image in the final asset. This image was only a placeholder.

Either way sounds good to me though!

ginnywood: Oh, this is an example, right, there would be no final asset. Should have just corrected in the doc—I’ll do that next time. Sorry for the confusion.

src/examples/text-inputs/FauxInput.tsx
hzhu: Text content has been added within the `FauxInput`.

src/nav/components.yml
hzhu: Updated page title from "Text inputs" to "Input" and IA.

src/pages/components/text-inputs.mdx
jzempel: If the majority case is that a bare input has no label, then maybe we should show that? But let's do it the right way. There is no need for `aria-label`. The technique would be to use `<Label hidden={true}>` – allowing the VO to read out the associated label when the bare input is focused.

src/pages/components/text-inputs.mdx
jzempel: Vertical margining with the current grid isn't great. Ideally there would be no margin when side-by-side, and then only interior margining when the elements are stacked.

src/examples/text-inputs/Size.tsx
hzhu: Removed 👍

src/pages/components/text-inputs.mdx
hzhu: Awesome sounds good, removed the `aria-label` copy.

src/pages/components/text-inputs.mdx
hzhu: Update: Spoke with @rossmoody and we've punted this section.

src/pages/components/text-inputs.mdx
hzhu: I think the guidelines could be flushed out after v1 (if there are guidelines). This copy has been revised.

hzhu:
> Most of the examples have different widths, even for the same kind of input. This makes the page less cohesive. Is there a reason to make them different widths, or can we pick a consistent width?

No particular reason — the elements are rendered as they are so take up default spacing which turns out to be different depending on how long the label/hint text is and how many other elements are around each field. 

A consistent width could be added to field inputs not rendered in a `Grid`. The tradeoff would be that the example includes website specific styles.

Is there are particular size that should be used?

> What is determining the order of the examples?

Originally the examples followed the order of the Google doc. However, the Google doc was updated yesterday. How should order be determined? Is it by the presence of placeholder text?

@allisonacs

src/pages/components/text-inputs.mdx
```diff
@@ -0,0 +1,199 @@
...
+      <li>For read-only objects, use text</li>
+    </ul>
+  </Misuse>
+</Usage>
+
+## Examples
+
+### Input
+
+A text input is a form field that lets the user enter text.
```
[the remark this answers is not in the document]
hzhu: Update: this copy has been revised.

src/pages/components/text-inputs.mdx
```diff
@@ -0,0 +1,199 @@
...
+  <FauxInput />
+</CodeExample>
+
+### Regular weight label
+
+Text input labels are emphasized by default. They can be regular weight using the `isRegular`
+property.
+
+Regular labels are not recommended for text inputs, as they flatten visual hierarchy making
+scanning more difficult.
```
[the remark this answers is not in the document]
hzhu: Update: This example has been removed from the PR.

allisonacs:
> Originally the examples followed the order of the Google doc. However, the Google doc was updated yesterday. How should order be determined? Is it by the presence of placeholder text?

I'm hoping @vsri1137 can determine a standard ordering for some of these things, but in the meantime, this is the order I feel makes the most sense in terms of how often some of these things are used. 

1. Input (default)
2. Textarea
3. Hint
4. Size
5. Validation
6. MediaInput
7. Focus inset
8. Placeholder
9. Regular weight label
10. FauxInput
11. Bare
12. Inline
13. Disabled

src/pages/components/text-inputs.mdx
hzhu: @rossmoody updated the lines so they are a single sentence. 

> Also the example feels like it's missing the described other component flush with the text input.

Could you clarify the issue and the change that is needed? I'm unsure which other component should be flushed with the text input.

allisonacs: Most common scenario for the inset focus is to butt a text input up directly against a button. For example, a media input with a search icon that sits flush to the left of a button with a label of "Search" (this would utilize the chevron button I believe, which is used when creating the split button, but i could be wrong about that).

src/pages/components/text-inputs.mdx
hzhu: > The responsive resizing on this example feels off. Could we do something more like what we see on menus/selects?

@allisonacs yeah it does feel off, I've made some adjustments. Could you let me know what you think? For small screen sizes the inputs should be stacked now.

src/examples/forms/inputs/Validation.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import React from 'react';
+import styled from 'styled-components';
+import { Grid, Row, Col } from '@zendeskgarden/react-grid';
+import { Field, Label, Input, Message } from '@zendeskgarden/react-forms';
+
+const StyledRow = styled(Row)`
+  @media (max-width: ${props => props.theme.breakpoints.md}) {
+    display: flex;
+    flex-direction: column;
```
jzempel, line 16: A `Row` is already `display: flex` and `flex-direction` should not be used with Garden's grid. Use props on the `Col`, instead. i.e. `<Col sm>` to stack at the small breakpoint.

So, in order to get vertical spacing, this becomes (replaces the 2nd and 3rd columns):

```js
const StyledCol = styled(Col)`
  @media (max-width: ${p => p.theme.breakpoints.sm}) {
    margin-top: ${p => p.theme.space.sm};
  }
`;
```

...but at this point, it's questionable for me whether this is a pattern we want to push through the examples. If so, let's tighten-up throughout. If not, then revert to using `<Col sm>` when columns need to stack on mobile.

src/examples/forms/inputs/Validation.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    > div:not(:first-child) {
```
jzempel, line 18: Per previous comment, this should not be needed. But, in cases where it is, always prefer to reference a component when available, i.e. `${Col}:not(:first-child)`.

src/pages/components/inputs.mdx
```diff
@@ -0,0 +1,165 @@
...
+package: '@zendeskgarden/react-forms'
+propSheets:
+  - forms/src/elements/FauxInput.tsx
+  - forms/src/elements/common/Field.tsx
+  - forms/src/elements/common/Hint.tsx
+  - forms/src/elements/Input.tsx
+  - forms/src/elements/common/Label.tsx
+  - forms/src/elements/MediaInput.tsx
+  - forms/src/elements/common/Message.tsx
+  - forms/src/elements/Textarea.tsx
```
jzempel, line 14: In other places, we've gone with component priority ordering rather than alphabetical. Please align with @rossmoody.

src/examples/forms/inputs/Validation.tsx
```diff
@@ -0,0 +1,47 @@
...
+const Example = () => (
+  <Grid>
+    <Row>
+      <StyledCol sm={4}>
+        <Field>
+          <Label>Plant</Label>
+          <Input validation="success" />
+          <Message validation="success">A cactus is a beautiful plant</Message>
+        </Field>
+      </StyledCol>
```
jzempel, line 28: In order to avoid unwanted margining, this one would just be a `Col`.

hzhu:
> Most of the examples have different widths, even for the same kind of input. This makes the page less cohesive. Is there a reason to make them different widths, or can we pick a consistent width?

The widths have size are rendered out of the box as-is. Styling could be applied to the examples.

jzempel:
> No extra styles were used. Styling could be applied to the examples. What widths should the examples use?

@hzhu let's talk about this today. Looks like the code example wrapper div is `display: flex`. I'm not convinced that's right – might be best to leave it as default `display: block`. That allows all components to layout as intended. In the case of fields, they would become full width as they are block elements.

Second, it may be desirable to apply `Grid` to all examples. In the case of a single field, we might use something like:

```
const Example = () => (
  <Grid>
    <Row justifyContent="center">
      <Col sm={6}>
        <Field>
          <Label>Plant</Label>
          <Input />
        </Field>
      </Col>
    </Row>
  </Grid>
);
```

...to obtain layout consistency on the page.

hzhu:
Merging to `master` ✅ . There are two follow up PRs that will be published soon:

1) Make changes to copy once @vsri1137 finishes tidying up copy this week.
2) Make adjustments to spacing and widths once [flex styling](https://github.com/zendeskgarden/website/blob/master/src/components/MarkdownProvider/components/CodeExample/index.tsx#L72) is removed from the `CodeExample` containers.

PR #63, opened 2020-06-11
chore: overview links

rossmoody:
## Description

This is an approach for the temporary overview page links to address navigating when the sidebar disappears. I was thinking that borrowing a similar approach from the sidebar navigation for the overview pages might be nice for programmatically showing the correct links in each section. This way we don't have to manually keep track of IA changes on each overview page. 

## Checklist

### DIFF
--- a/src/components/MarkdownProvider/components/OverviewLinks.tsx
+++ b/src/components/MarkdownProvider/components/OverviewLinks.tsx
@@ -1,0 +1,60 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { ISidebarSection } from '../../../layouts/Sidebar';
+import { LG, UnorderedList, MD } from '@zendeskgarden/react-typography';
+import { Link } from '../../../layouts/Root/components/StyledNavigationLink';
+import styled from 'styled-components';
+
+const StyledUnorderedList = styled(UnorderedList)`
+  margin: 0 0 ${p => p.theme.space.sm} 0;
+`;
+
+const StyledListItem = styled(UnorderedList.Item)`
+  list-style: none;
+`;
+
+export const OverviewLinks: React.FC<{ nav: ISidebarSection[] }> = ({ nav }) => {
+  const content = nav.map(section => (
+    <StyledUnorderedList key={`${section.title}`}>
+      <LG isBold>{section.title}</LG>
+      {section.items && (
+        <StyledUnorderedList>
+          {section.items?.map(group => {
+            if (group.items) {
+              return (
+                <StyledListItem>
+                  <MD>{group.title}</MD>
+                  <UnorderedList>
+                    <UnorderedList.Item>
+                      {group.items.map(child => {
+                        return (
+                          <UnorderedList.Item key={`${child.title}`}>
+                            <Link to={`${child.id}`}>{child.title}</Link>
+                          </UnorderedList.Item>
+                        );
+                      })}
+                    </UnorderedList.Item>
+                  </UnorderedList>
+                </StyledListItem>
+              );
+            }
+
+            return (
+              <StyledListItem key={`${group.title}`}>
+                <Link to={`${group.id}`}>{group.title}</Link>
+              </StyledListItem>
+            );
+          })}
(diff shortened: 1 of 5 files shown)


rossmoody:
@allisonacs @hzhu @jzempel @austin94 This is in draft mode and comparing against the `website-ia` branch that might shift, but I'd be curious for thoughts on what's going on here. Both in the code approach and also if we would be ok with the visual aesthetic as a temporary solve until the overview page can be designed and implemented correctly.

jzempel:
I don't have a big problem with it as a temporary hack. My only nit is to make sure those unordered lists get a `type="disc"` prop so that they match other usages throughout the site.

allisonacs:
Yea, no real issue as a temp solution. If you can throw it into a column structure so we can avoid it feeling so much like a link page, all the better. But as a temp solution I'm good with it. +1 to @jzempel's comment re: disc bullets.

src/components/MarkdownProvider/components/OverviewLinks.tsx
```diff
@@ -0,0 +1,58 @@
...
+  margin: 0 0 ${p => p.theme.space.sm} 0;
+`;
+
+const StyledListItem = styled(UnorderedList.Item)`
+  list-style: none;
+`;
+
+export const OverviewLinks: React.FC<{ nav: ISidebarSection[] }> = ({ nav }) => {
+  const content = nav.map(section => (
+    <StyledUnorderedList key={`${section.title}`}>
```
hzhu, line 24: Since this string template doesn't involve additional string values this could be updated to not use string template. It would look like `key={section.title}`. Other string template values in this file could be updated in the same way.

src/pages/components/index.mdx
```diff
@@ -1,12 +1,13 @@
 ---
 title: Overview
 description: >
-  A standard overview
+  Development instructions for building effective user interfaces.
 ---
 
-## Introduction
+import { OverviewLinks } from '../../components/MarkdownProvider';
```
hzhu, line 7: I am seeing some new warnings both locally and in the CI build 👀 

Screen Shot 2020-06-15 at 11 45 36 AM

I think these imports may not be needed, and once removed seems to address the warning logs.

src/components/MarkdownProvider/components/OverviewLinks.tsx
rossmoody: Good spot. I updated all the `key` instances.

src/pages/components/index.mdx
rossmoody: TY 👍

src/components/MarkdownProvider/components/OverviewLinks.tsx
```diff
@@ -0,0 +1,58 @@
...
+                        </UnorderedList.Item>
+                      );
+                    })}
+                  </UnorderedList>
+                </StyledListItem>
+              );
+            }
+
+            return (
+              <StyledListItem key={`${group.title}`}>
```
hzhu, line 47: Whoops, I think a couple of extra string templates exists.

src/components/MarkdownProvider/components/OverviewLinks.tsx
```diff
@@ -0,0 +1,58 @@
...
+          {section.items?.map(group => {
+            if (group.items) {
+              return (
+                <StyledListItem key={group.title}>
+                  {group.title}
+                  <UnorderedList type="disc">
+                    {group.items.map(child => {
+                      return (
+                        <UnorderedList.Item key={child.title}>
+                          <Link to={`${child.id}`}>{child.title}</Link>
```
hzhu, line 37: Extra string template exist on `to` prop.

rossmoody: VSCode gets upset with me if I remove the literals on the Gatsby `to`: https://share.getcloudapp.com/RBuqvmOv Any idea what might be causing that?

rossmoody: Also, if you run `yarn clean` do you still get those export warnings? I'm not getting any in my console and I'm not sure why.

hzhu: > Also, if you run yarn clean do you still get those export warnings? I'm not getting any in my console and I'm not sure why.

I should have clarified, the logs from my latest comment are from the CI build. The warnings don't appear locally anymore. Could be a stale CircleCI cache.

PR #68, opened 2020-06-16
feat!: new PropSheet component

jzempel:
## Description

Note this is a BREAKING CHANGE. See below for [detail](#detail). The following enumerates the list of issues/improvements addressed:

- Per-component `<PropSheet>`
  - allows component name `<H3>` MDX to appear in API nav
  - allows author to provide component description (`react-components` descriptions do not make the quality cut required for the website)
  - allows interspersed `<H4>` content – static accessor components (i.e. `Button.StartIcon`), for example
- "Required" tags added to the "Default" column. See the `Multiselect` API for example. (Design [discussion](#discussion) needed.)
- Function-style (noisy) types removed from table flow and added as tooltips. Design [discussion](#discussion) needed.
- Remove props marked with `@ignore`
- Using `shouldRemoveUndefinedFromOptional` docgen config in order to remove ` | undefined` (noisy) strings from optional types
- Surround string literal default values with single quotes
- Provide a more intuitive component data structure that we control instead of accepting `ComponentDoc` straight from docgen:

```
{
  name: string
  description: string
  props: {
    description: string
    defaultValue: string
    required: boolean
    type: string
  }
}
```

## Detail

Upon merge, any in-process PRs will need to make the following changes:

1. Frontmatter: rename from `propSheets` to `components`.
1. `<Configuration>`: replace `propSheets` prop with `components={props.data.mdx.components}`
1. `<PropSheets>` deleted. Replace with:

```mdx
### ComponentX

<PropSheet components={props.data.mdx.components} componentName="componentX" />
```

## Discussion

Offline design discussions needed for the following:

- What to be done for component descriptions?
- How to handle "Required" prop values?
- How to handle "func" prop types?
- Doclet tags (`@param`, `@return`, etc) still need to be stripped from prop descriptions, added to the `components` data, and handled in the design of the prop sheets table.

## Checklist


src/pages/components/select.mdx
```diff
@@ -46,7 +46,7 @@ import SelectDefaultCode from '!!raw-loader!../../examples/dropdowns/SelectDefau
   <SelectDefault />
 </CodeExample>
 
-### Hint
+### Hint copy
```
jzempel, line 49: @rossmoody this was the copy update we discussed in order to avoid a same-level nav collision with the `API > Hint` below. Depending on what you figure out with @vsri1137, we can restore this and turn off the `markdownlint` rule (as our heading parser handles duplicates) or align with copy standards.

src/pages/components/menu.mdx
```diff
@@ -132,11 +131,57 @@ import MenuMediaCode from '!!raw-loader!../../examples/dropdowns/MenuMedia.tsx';
...
+
+### HeaderItem
+
+<PropSheet components={props.data.mdx.components} componentName="headerItem" />
+
+### Item
+
+<PropSheet components={props.data.mdx.components} componentName="item" />
+
+### ItemMeta
```
hzhu, line 162: Some components don't have additional props besides their native backing element attributes, for example [ItemMeta](https://5ee914ebd759f501f95844d8--zendeskgarden.netlify.app/components/menu#itemmeta), MediaBody, MediaFigure, Separator, etc. Should the copy eventually be something along the lines of "Accepts all `<div>` props"?

Screen Shot 2020-06-16 at 12 53 01 PM

src/pages/components/menu.mdx
```diff
@@ -132,11 +131,57 @@ import MenuMediaCode from '!!raw-loader!../../examples/dropdowns/MenuMedia.tsx';
...
+
+<PropSheet components={props.data.mdx.components} componentName="itemMeta" />
```
hzhu, line 164: Should component names be PascalCased?

src/pages/components/menu.mdx
jzempel: I really hope not re: "Accepts all `<xxx>` props". We need to DRY up redundant copy and add it to the component overview page. But per **Discussion** mentioned in PR description, the lacking description paragraphs need design consideration. In the end, every single component needs a helpful high-level description.

src/pages/components/menu.mdx
jzempel: It doesn't matter – they can be any-cased. I just thought pascalCased looked best.

src/pages/components/select.mdx
rossmoody: This seems like a great temporary band-aid to me.

src/pages/components/menu.mdx
rossmoody: I think this is a great opportunity for us to write one or two line utility descriptors for these components (not now, but soon). I personally find myself looking at these components and then scrolling up to examples to determine how they should be used. Could be nice to try to explain the utility here to help eliminate that back and forth.

allisonacs:
I'm seeing a few instances where it looks like there should be a prop table (by way of the heading), but none appears. Is this expected? 

image
(example from multiselect page)

vsri1137:
test comment

jzempel:
> I'm seeing a few instances where it looks like there should be a prop table (by way of the heading), but none appears. Is this expected?

Yes, @allisonacs, this is expected. The `<PropSheet>` component exists, but doesn't render anything if there are no props. If props are added in the future, these components will populate. In the mean time, these are markers for where we should add a description paragraph (TBD).

src/pages/components/inputs.mdx
```diff
@@ -81,7 +81,7 @@ import IconsCode from '!!raw-loader!../../examples/forms/inputs/MediaInput.tsx';
   <Icons />
 </CodeExample>
 
-### FauxInput
+### Faux input
```
jzempel, line 84: @rossmoody just a ping that there are more copy updates needed with the latest merge from `master`.

PR #69, opened 2020-06-17
chore: update avatar component page

rossmoody:
This PR introduces string revisions from UXCS to the Avatar component page. 

## Description

We've updated the content for this component page. This is meant to be a prototype for other component pages that follow. You can see some of the rationale for the direction being presented in the Avatar Google doc. 

## Checklist

### DIFF
--- a/src/pages/components/avatar.mdx
+++ b/src/pages/components/avatar.mdx
@@ -71,14 +71,10 @@ user distinguish between the two when scanning a page.
 
 <BestPractice>
   <Do imageSource={props.data.avatarShapeCircle.children[0].publicURL}>
-    <ul>
-      <li>Use a circular Avatar for people.</li>
-    </ul>
+    <p>Use a circular Avatar for people.</p>
   </Do>
   <Do imageSource={props.data.avatarShapeSquare.children[0].publicURL}>
-    <ul>
-      <li>Use a rounded square Avatar for objects, brands, or Zendesk products.</li>
-    </ul>
+    <p>Use a rounded square Avatar for objects, brands, or Zendesk products.</p>
   </Do>
 </BestPractice>
 
--- a/src/pages/components/avatar.mdx
+++ b/src/pages/components/avatar.mdx
@@ -1,7 +1,7 @@
 ---
 title: Avatar
 description: >
-  An avatar is a visual way to represent a person or brand in the product. They can display text,
+  An Avatar is a visual way to represent a person or brand in the product. They can display text,
   icons, or images.
 package: '@zendeskgarden/react-avatars'
 propSheets:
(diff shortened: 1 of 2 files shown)


src/pages/components/avatar.mdx
```diff
@@ -53,8 +53,7 @@ import AvatarTypeCode from '!!raw-loader!../../examples/avatar/AvatarType.tsx';
 
 ### Status
 
-A status ring can be applied to the outside of an Avatar to communicate a user is away, available,
-or active.
+A colored ring outside the Avatar indicates a user's status. The options are away, available, or active.
```
jzempel, line 56: re: "user's"... What is the opinion re: https://en.wikipedia.org/wiki/Apostrophe – straight or curly? In the same vein, should we be checking for balanced quotes or other rules? If so, we might want to incorporate a tool like https://textlint.github.io/ into the website build.

src/pages/components/avatar.mdx
```diff
@@ -63,20 +62,23 @@ import AvatarStatusCode from '!!raw-loader!../../examples/avatar/AvatarStatus.ts
...
-    <p>Use a circular shape to represent a person.</p>
+    <ul>
+      <li>Use a circular Avatar for people.</li>
+    </ul>
   </Do>
   <Do imageSource={props.data.avatarShapeSquare.children[0].publicURL}>
-    <p>Use the rounded square shape to represent objects, brands, or products.</p>
+    <ul>
+      <li>Use a rounded square Avatar for objects, brands, or Zendesk products.</li>
+    </ul>
```
jzempel, line 81: Why were these changes made? When the information is not a list, a `<p>` is the more semantic HTML.

rossmoody:
@jzempel 
1. I will add the words to be mindful of for conflicts to our component template doc. 
2. I'd love @vsri1137's help on the prop sheet writing refinement we've done so far. Might be a nice after-MVP ticket. We have a template started but it would be smart to get UXCS input before going through all the components.

src/pages/components/avatar.mdx
vsri1137: > How will the single-word "How to use it" headings scale to other pages? Avatar is a simple page. Consider a more complex page such as Inputs where example headings need to avoid conflicting with API component names.

You're referring to those those single-word subheadings like "Default" "Size" "Disabled" etc? If so, I'd first want to know how we think of those. Are those "features" of the component? Settings?

jzempel: Above I am talking about the single word headings you listed. I'll let @rossmoody answer to how we think about them.

jzempel: @vsri1137 I'd also like to hear your thoughts on the single vs. curly quote situation. We don't need to hold-up this PR, but there are automated linting solutions we can introduce to auto-fix some of these nits in the future if desired.

PR #75, opened 2020-06-19
feat: add checkbox component page

hzhu:
## Description

This pull request adds the Checkbox component page.

## Checklist

### DIFF
--- a/src/examples/forms/checkbox/Default.tsx
+++ b/src/examples/forms/checkbox/Default.tsx
@@ -1,0 +1,29 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Field, Label, Checkbox } from '@zendeskgarden/react-forms';
+import { Row, Col } from '@zendeskgarden/react-grid';
+
+const Example = () => (
+  <Row justifyContent="center">
+    <Col size="auto">
+      <Field>
+        <Checkbox defaultChecked>
+          <Label>Annual</Label>
+        </Checkbox>
+      </Field>
+      <Field>
+        <Checkbox>
+          <Label>Perennial</Label>
+        </Checkbox>
+      </Field>
+    </Col>
+  </Row>
+);
+
+export default Example;
--- a/src/examples/forms/checkbox/Default.tsx
+++ b/src/examples/forms/checkbox/Default.tsx
@@ -5,25 +5,30 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 import { Field, Label, Checkbox } from '@zendeskgarden/react-forms';
 import { Row, Col } from '@zendeskgarden/react-grid';
 
-const Example = () => (
-  <Row justifyContent="center">
-    <Col size="auto">
-      <Field>
-        <Checkbox defaultChecked>
-          <Label>Annual</Label>
-        </Checkbox>
-      </Field>
-      <Field>
-        <Checkbox>
-          <Label>Perennial</Label>
-        </Checkbox>
-      </Field>
-    </Col>
-  </Row>
-);
+const Example = () => {
+  const [annual, setAnnual] = useState(true);
+  const [perennial, setPerennial] = useState(false);
+
+  return (
+    <Row justifyContent="center">
+      <Col size="auto">
+        <Field>
+          <Checkbox checked={annual} onChange={() => setAnnual(!annual)}>
+            <Label>Annual</Label>
+          </Checkbox>
+        </Field>
+        <Field>
+          <Checkbox checked={perennial} onChange={() => setPerennial(!perennial)}>
+            <Label>Perennial</Label>
+          </Checkbox>
+        </Field>
+      </Col>
+    </Row>
+  );
+};
 
 export default Example;
--- a/src/examples/forms/checkbox/Indeterminate.tsx
+++ b/src/examples/forms/checkbox/Indeterminate.tsx
@@ -1,0 +1,101 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useReducer } from 'react';
+import styled from 'styled-components';
+import { Field, Label, Checkbox } from '@zendeskgarden/react-forms';
+import { Row, Col } from '@zendeskgarden/react-grid';
+
+interface ISelectPerennial {
+  type: 'perennial';
+}
+
+interface ISelectAnnual {
+  type: 'annual';
+}
+
+interface ISelectAll {
+  type: 'all';
+}
+
+interface ISelectNone {
+  type: 'none';
+}
+
+const initialState = { annual: true, perennial: false };
+
+type ActionTypes = ISelectPerennial | ISelectAnnual | ISelectAll | ISelectNone;
+
+type ReducerType = (state: typeof initialState, action: ActionTypes) => typeof initialState;
+
+const reducer: ReducerType = (state, action) => {
+  switch (action.type) {
+    case 'all':
+      return { annual: true, perennial: true };
+    case 'none':
+      return { annual: false, perennial: false };
+    case 'annual':
+      return { ...state, annual: !state.annual };
+    case 'perennial':
+      return { ...state, perennial: !state.perennial };
+    default:
+      return state;
+  }
+};
+
+const StyledGroup = styled(Field)`
+  margin-${props => (props.theme.rtl ? 'right' : 'left')}: ${props => props.theme.space.base * 6}px;
+`;
+
+const Example = () => {
+  const [state, dispatch] = useReducer(reducer, initialState);
+  const { annual, perennial } = state;
+
+  const onParentChange = () => {
+    if (perennial || annual) {
+      dispatch({ type: 'all' });
+    }
+
+    if (!perennial || !annual) {
+      dispatch({ type: 'all' });
+    }
+
+    if (perennial && annual) {
+      dispatch({ type: 'none' });
+    }
+  };
+
+  return (
+    <Row justifyContent="center">
+      <Col size="auto">
+        <Field>
+          <Checkbox
+            onChange={onParentChange}
+            checked={perennial && annual}
+            indeterminate={!perennial && !annual ? false : !perennial || !annual}
+          >
+            <Label>Growth types</Label>
+          </Checkbox>
+        </Field>
+        <StyledGroup>
+          <Field>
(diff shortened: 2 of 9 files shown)


src/examples/forms/checkbox/Default.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Field, Label, Checkbox } from '@zendeskgarden/react-forms';
+import { Row, Col } from '@zendeskgarden/react-grid';
+
+const Example = () => (
+  <Row justifyContent="center">
+    <Col size="auto">
+      <Field>
+        <Checkbox defaultChecked>
+          <Label>Annual</Label>
+        </Checkbox>
```
austin94, line 18: Should we show these as controlled inputs for the default example? Otherwise it might not be obvious that they map to a native `<input />`

src/pages/components/checkbox.mdx
```diff
@@ -1,15 +1,183 @@
...
+import DefaultCode from '!!raw-loader!../../examples/forms/checkbox/Default.tsx';
+
+<CodeExample code={DefaultCode}>
+  <Default />
+</CodeExample>
+
+### Indeterminate
+
+If a checkbox has child checkboxes, it should show an indeterminate state (reflected by a dash) when
+the child checkboxes contains both some checked and unchecked checkboxes.
```
austin94, line 48: > child checkboxes contains both some

Is this grammatically correct? Parent-child interactions can mean specific things on the dev side, specifically around DOM hierarchy.

Maybe a more correct wording would be "grouping".

hzhu: > Is this grammatically correct?

I'm not sure, but it sounds like the sentence could be improved a lot. I was hoping @vsri1137 could help check out the example and wordsmith that description 🤞🏽.

> Maybe a more correct wording would be "grouping".

I think the word "grouping" doesn't convey a (visual, not DOM) parent-child relationship. The term "child checkboxes" does convey that visual relationship. Apple HCI guidelines use [similar](https://developer.apple.com/design/human-interface-guidelines/macos/buttons/checkboxes/) terminology. Maybe there is another term that conveys a visual parent-children hierarchy?

src/examples/forms/checkbox/Default.tsx
hzhu: Hmm, I didn't think it was necessary to show the default example as a controlled component. 

How does a controlled example demonstrate that a native input backs the component? Is it because native attributes are passed, such as `checked`? `defaultChecked` is also a native input attribute.

If it's important that the audience knows that `<Checkbox>` is backed by a native input element, I think the doc should explicitly communicate that in the component description section (which is currently blank).

austin94: The majority of forms usage is "controlled" at this point. I think showing both a controlled and uncontrolled usage could be useful.

> How does a controlled example demonstrate that a native input backs the component

Specifically the `onChange` event shows the properties required to handle this. The `material-ui` [Checkbox example](https://material-ui.com/components/checkboxes/#checkbox) has a decent example of this. Probably not necessary for every example though.

jzempel: Current Garden also has an example. Won't speak to the quality :wink: https://garden.zendesk.com/react-components/forms/#advanced

src/pages/components/checkbox.mdx
allisonacs: @austin94 Given the context, is it likely that there would be confusion here around the use of parent-child? 

If so, perhaps something more accurate that grouping (which implies several options that are all at the same level) we could use ancestor-descendant. 

However, if there is little chance that someone reading this will mistake this for DOM usages, then I'd rather we stick with more well-known, less jargon-y words.

src/examples/forms/checkbox/Indeterminate.tsx
```diff
@@ -0,0 +1,101 @@
...
+          <Field>
+            <Checkbox checked={annual} onChange={() => dispatch({ type: 'annual' })}>
+              <Label>Annual</Label>
+            </Checkbox>
+          </Field>
+          <Field>
+            <Checkbox checked={perennial} onChange={() => dispatch({ type: 'perennial' })}>
+              <Label>Perennial</Label>
+            </Checkbox>
+          </Field>
```
allisonacs, line 94: Because the parent is bolded, stylistically, it will make more sense for these options to have regular weight labels.

src/pages/components/checkbox.mdx
```diff
@@ -1,15 +1,183 @@
...
+A Checkbox’s label is part of its touch target.
+
+import Default from '../../examples/forms/checkbox/Default.tsx';
+import DefaultCode from '!!raw-loader!../../examples/forms/checkbox/Default.tsx';
+
+<CodeExample code={DefaultCode}>
+  <Default />
+</CodeExample>
+
+### Indeterminate
```
allisonacs, line 45: Because I've suggested that the child checkboxes receive default weight labels, we should switch the order of the examples so that regular weight label example appear above the indeterminate example.

src/pages/components/checkbox.mdx
```diff
@@ -1,15 +1,183 @@
...
+<CodeExample code={RegularWeightLabelCode}>
+  <RegularWeightLabel />
+</CodeExample>
+
+### Hint text
+
+Hint adds additional context to the Checkbox label.
+
+import Hint from '../../examples/forms/checkbox/Hint.tsx';
+import HintCode from '!!raw-loader!../../examples/forms/checkbox/Hint.tsx';
```
allisonacs, line 73: 🙌 Nice job adding the extra space between these @hzhu thank you!

src/pages/components/checkbox.mdx
```diff
@@ -1,15 +1,183 @@
...
+    <ul>
+      <li>To select any number of options, including zero</li>
+      <li>As a standalone control to turn a single option on or off</li>
+    </ul>
+  </Use>
+  <Misuse>
+    <ul>
+      <li>
+        To limit the user to selecting exactly one option from a set, use{' '}
+        <a href="../radio">Radios</a>
```
jzempel, line 26: Invalid href. Also check "Radios" wording against "Radio" page. Copy might need to be something like "use the _Radio_ component"

src/pages/components/checkbox.mdx
```diff
@@ -1,15 +1,183 @@
...
+    <p>Use to select multiple options from a set of two or more.</p>
+    <p>
+      Use to turn a standalone setting on and off, as a favorable alternative to a{' '}
+      <a href="./toggle">Toggle</a>.
+    </p>
+  </Do>
+  <Dont imageSource={props.data.checkboxStandaloneOptionDont.children[0].publicURL}>
+    <p>
+      Use to select exactly one option from a set of two or more. Use <a href="./radio">Radios</a>
+      instead.
```
jzempel, line 128: Missing space. Also, similar comment re: "Radios" vs. "Radio" (defer to word folks).

hzhu: Fixed this invalid href and missing space. 

The text was copied from the reviewed Google Doc. Should it change?

cc: @vsri1137 @rossmoody

src/pages/components/checkbox.mdx
```diff
@@ -1,15 +1,183 @@
...
+    <p>Use to select multiple options from a set of two or more.</p>
+    <p>
+      Use to turn a standalone setting on and off, as a favorable alternative to a{' '}
+      <a href="./toggle">Toggle</a>.
+    </p>
+  </Do>
+  <Dont imageSource={props.data.checkboxStandaloneOptionDont.children[0].publicURL}>
+    <p>
+      Use to select exactly one option from a set of two or more. Use <a href="../radio">Radios</a>
+      &nbsp;instead.
```
jzempel, line 128: Please check the deployment. Now both `../radio` links are broken. Looks like we've used `/components/radio` style links in other locations. Also, seems we typically use `{' '}` rather than `&nbsp;` 🤷 

/cc @rossmoody

rossmoody: Oh, is this a manual fix because of the issue I referenced on SplitButtons with Prettier and MarkdownLint? If you update the url to "/components/radio" you might get lucky and have it break in a spot that passes.

rossmoody: For posterity, I wasn't adding {' '}, Prettier was adding it for me on `yarn format`

hzhu: A lint rule requires each line to be 100 characters maximum. Adjusting this sentence to have a 100 character limit gets the linting to pass during the commit hooks, but during the formatting phase, Prettier will insert `{'  '}` which pushes the character limit beyond 100.

jzempel: Sounds like the linting config should be tweaked.

jzempel: Just tested. The fact that `.prettierrc` and `.markdownlintrc` are equal to each other seems to be a problem. I'd recommend setting `.markdownlintrc` to `"line-length: 120"`.

hzhu: Awesome, sounds good! I'll update the linting config instead of manually adding `&nbsp;`.

hzhu: Awesome, sounds good! I'll update the linting config instead of manually adding `&nbsp;`.

Edit: Updated `.markdownlintrc` to `"line-length: 120"` so now Prettier rules play well with Markdown lint rules.

src/pages/components/checkbox.mdx
vsri1137: Proposed:

> If the children (or descendants) of a Checkbox have different states (some on, some off) the parent Checkbox will show an indeterminate state.

src/pages/components/checkbox.mdx
vsri1137: I was thinking of the current garden page which calls them `Radios`. Let's change to:

> To give the user a mutually exclusive choice use the [Radio] component instead

src/pages/components/checkbox.mdx
hzhu: Thanks V, that sounds great. Huge improvement!

src/pages/components/checkbox.mdx
hzhu: Updated 👍

hzhu:
@vsri1137 I updated the content with the most recent updates from the Checkbox Google Doc. Could you review the PR (and approve) if it looks good to go?

src/pages/components/checkbox.mdx
```diff
@@ -1,15 +1,184 @@
...
+  - forms/src/elements/common/Hint.tsx
+  - forms/src/elements/common/Label.tsx
+  - forms/src/elements/common/Message.tsx
 ---
 
-Content here
+<Usage>
+  <Use>
+    <ul>
+      <li>To let users compare options from a list and select all, any, or none of those items.</li>
```
vsri1137, line 17: Line 17 shouldn't have closing punctuation:

`<li>To let users compare options from a list and select all, any, or none of those items</li>`

hzhu: Updated, thanks V!

PR #110, opened 2020-07-09
fix: allow table of contents to scroll

hzhu:
## Description

Users should be able to scroll through the Table of Contents (on Desktop sizes) with lots of content (e.g. [Word List](https://zendeskgarden.netlify.app/content/word-list)). This PR updates the `TOC` component so that long Tabe of Contents can be scrolled.

## Checklist

### DIFF
--- a/src/layouts/Titled/components/TOC.tsx
+++ b/src/layouts/Titled/components/TOC.tsx
@@ -121,7 +121,7 @@ export const TOC: React.FC<{ data: IHeading[] }> = ({ data }) => {
         position: sticky;
         top: ${p => p.theme.space.lg};
         margin-left: ${p => p.theme.space.base * 15}px;
-        max-height: calc(100vh - 64px);
+        max-height: calc(100vh - ${p => p.theme.space.base * 16}px);
         overflow-y: scroll;
         -ms-overflow-style: scrollbar;
       `}
--- a/src/layouts/Titled/components/TOC.tsx
+++ b/src/layouts/Titled/components/TOC.tsx
@@ -121,6 +121,8 @@ export const TOC: React.FC<{ data: IHeading[] }> = ({ data }) => {
         position: sticky;
         top: ${p => p.theme.space.lg};
         margin-left: ${p => p.theme.space.base * 15}px;
+        max-height: ${OFFSET}px;
+        overflow-y: scroll;
       `}
     >
       <StyledSectionHeader
@@ -133,7 +135,6 @@ export const TOC: React.FC<{ data: IHeading[] }> = ({ data }) => {
       </StyledSectionHeader>
       <ul
         css={css`
-          margin-left: -${p => p.theme.borderWidths.sm};
           border-left: ${p => p.theme.borders.sm} ${p => getColor('grey', 200, p.theme)};
         `}
       >
--- a/src/layouts/Titled/components/TOC.tsx
+++ b/src/layouts/Titled/components/TOC.tsx
@@ -122,6 +122,7 @@ export const TOC: React.FC<{ data: IHeading[] }> = ({ data }) => {
         position: sticky;
         top: ${p => p.theme.space.lg};
         margin-left: ${p => p.theme.space.base * 15}px;
+        padding-right: ${p => p.theme.space.base * 4}px;
         max-height: calc(100vh - ${p => math(`${p.theme.space.lg} * 2`)});
         overflow-y: scroll;
         -ms-overflow-style: scrollbar;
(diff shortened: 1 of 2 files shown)


src/layouts/Titled/components/TOC.tsx
```diff
@@ -121,6 +121,9 @@ export const TOC: React.FC<{ data: IHeading[] }> = ({ data }) => {
         position: sticky;
         top: ${p => p.theme.space.lg};
         margin-left: ${p => p.theme.space.base * 15}px;
+        max-height: ${OFFSET}px;
```
jzempel, line 124: This wants to be a `calc` of something like `(100vh - 64px)` - where 64 pixels is the existing top and bottom 32px of padding. Also, the width might need some design love in order to better accommodate the scrollbar size.

allisonacs:
Looks like this one could be tougher than we hoped. Take a look at this video. https://share.getcloudapp.com/geuw71pK

Normally, I'd expect the active TOC/highlighted element to be visible. With the new implementation, it will need to auto scroll to see the active item.

jzempel:
@allisonacs given the tech we're using to track position, I think this behavior can be considered non-optimal yet normal and somewhat expected. Please see https://www.gatsbyjs.org/docs/gatsby-image/ (reduce vertical browser height). The same situation occurs, and I don't find it totally off-putting. I know exactly how to scroll in order to obtain context, if needed.

src/layouts/Titled/components/TOC.tsx
hzhu: Since the width of the TOC list isn't explicit, should `padding-right` be added to the TOC list to account for the scrollbar size? I think `4px` could work. Any suggestions come to mind? What do ya'll think? 

@allisonacs @rossmoody

jzempel: We have research and helpful guidance under a now-deprecated CSS component – https://garden.zendesk.com/css-components/scrollbars/ – to understand that standard "rail" width for a scrollbar is 15px. In this case, I think we can use our base-4 and apply an extra 16px of space.

allisonacs: @hzhu I might do something more like `20px`. The default width of a scroll bar on macos is 7px, but it's 17 on windows.

src/layouts/Titled/components/TOC.tsx
```diff
@@ -121,6 +121,9 @@ export const TOC: React.FC<{ data: IHeading[] }> = ({ data }) => {
         position: sticky;
         top: ${p => p.theme.space.lg};
         margin-left: ${p => p.theme.space.base * 15}px;
+        max-height: calc(100vh - 64px);
```
jzempel, line 124: Please use theme values and calculations when they are available. In this case the 64px is 2X the value of `top` coming from `p.theme.space.lg`.

src/layouts/Titled/components/TOC.tsx
```diff
@@ -121,6 +121,9 @@ export const TOC: React.FC<{ data: IHeading[] }> = ({ data }) => {
...
+        overflow-y: scroll;
+        -ms-overflow-style: scrollbar;
```
jzempel, line 126: Please follow the advice provided in the description of https://garden.zendesk.com/css-components/scrollbars/ and use `-ms-overflow-style: -ms-autohiding-scrollbar` rather than `scrollbar`. Instead of adding here, apply this to the [`GlobalStyling`](https://github.com/zendeskgarden/website/blob/master/src/layouts/Root/index.tsx) as:

```css
* {
  -ms-overflow-style: -ms-autohiding-scrollbar;
}
```

src/layouts/Titled/components/TOC.tsx
hzhu: Would `p.theme.space.base * 16` be okay? `p.theme.space.lg` is a string value that includes "px".

jzempel: You have the same go-to's we use all over `react-components` such as using polished `math`.

src/layouts/Titled/components/TOC.tsx
hzhu: Awesome thanks for the size suggestions. I went with a larger, `16px` space.

src/layouts/Titled/components/TOC.tsx
```diff
@@ -121,6 +122,10 @@ export const TOC: React.FC<{ data: IHeading[] }> = ({ data }) => {
         position: sticky;
         top: ${p => p.theme.space.lg};
         margin-left: ${p => p.theme.space.base * 15}px;
+        padding-right: ${p => p.theme.space.base * 4}px;
```
jzempel, line 125: [nit] the preference would be to use `p.theme.space.md` here.

src/layouts/Titled/components/TOC.tsx
jzempel: @hzhu looks like you forgot to remove this line here.

PR #112, opened 2020-07-10
fix: use hr for visual presentation

hzhu:
## Description

VoiceOver reads aloud horizontal rules and their interactions, but the horizontal rules are used for visual presentation alone. This affects all component pages and the "Word List" page.

## Detail

A `role="presentation"` has been added to the horizontal rules.

**before**:

**after**:
Horizontal rules and their interactions are no longer read aloud by VoiceOver.

## Checklist

- [ ] ~:black_nib: copy updates are approved (add the content strategist as a reviewer)~

### DIFF
--- a/src/components/MarkdownProvider/components/Typography.tsx
+++ b/src/components/MarkdownProvider/components/Typography.tsx
@@ -57,7 +57,9 @@ export const StyledBlockquote = styled(LG).attrs({ tag: 'blockquote' })`
   font-size: ${p => p.theme.space.base * 4}px;
 `;
 
-export const StyledHr = styled.hr`
+export const StyledHr = styled.hr.attrs({
+  role: 'presentation'
+})`
   margin: ${p => p.theme.space.md} 0;
   border-top: ${p => p.theme.borders.sm} ${p => getColor('grey', 200, p.theme)};
 `;


hzhu:
I did some post PR research and found:

> The hr element represents a paragraph-level thematic break, e.g., a scene change in a story, or a transition to another topic within a section of a reference book.

The `<hr>` also has an implicit role of "separator". I think this means the current usage of `<hr>` is appropriate. However it seems weird that VoiceOver read aloud to indicate to the user that it (the horizontal rule) can be interactable:

> You are currently on a horizontal separator. This item is dimmed. To start interacting with the splitter, press Control-Option-Shift-Down-Arrow.

Following the steps outlined by VoiceOver, interacting with the splitter leads the user down a path where they can "change the splitter position" which can be increased or decreased. Regardless of which command used, the position always remains at 0%.

Since there isn't sufficient research to backing this change, I'm can close this PR.

jzempel:
I'm curious how you were able to nav to the HR. I don't have a VO config (or my keyboard-fu is weak) that allows me to get to it without clicking.

hzhu:
I believe VO keys can be used without a config (as long as VoiceOver is turned on). The VoiceOver commands are entered by holding down the Control and Option keys together, along with one or more other keys. In this case, I held down `Command + Option` and used the arrow keys to navigate the page, and then into the splitter.

jzempel:
hmm, that doesn't happen for me – must be something different between our configurations.

PR #116, opened 2020-07-15
chore: update syntax highlighting

jzempel:
## Description

The goal here was to get as close as possible to CodeSandbox syntax highlighting so that the transition between website and sandbox would feel seamless. Without swapping out the underlying `react-syntax-highlighter` for `prism-react-renderer` [used by CodeSandbox](https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/app/components/Markdown/Code.tsx) we can get pretty close. Fwiw, `prism-react-renderer` has a richer set of tokens that it identifies on parse.

For starters, I moved the `SyntaxHighlighter` over to `prism` and set the language to `tsx`. This had the immediate affect of balancing problematic JSX highlighting.

Then I used the [themeFromVSCode](https://github.com/FormidableLabs/prism-react-renderer/tree/master/tools/themeFromVsCode) converter to obtain a prism config that (supposedly) matches VSCode Default Dark + (the default theme used-ish by CodeSandbox). I compared that to the [vsdark styles](https://github.com/conorhastings/react-syntax-highlighter/blob/master/src/styles/prism/vs-dark.js) added to `react-syntax-highlighter` a while ago but not yet released to npm.

Frankly, all this gave me was a bunch of tokens that weren't quite color-aligned with CodeSandbox. So I went one-by-one and tried to come up with a matching hex color config. The end result is this PR.

## Detail

@allisonacs the outer wrapper was updated to use Garden monospace font-family, font-sizing, and line-height. I applied `theme.space.sm` (12px) padding all around.

## Checklist

- [ ] :zap: ~audited via [web.dev](https://web.dev/measure/) for performance and SEO~


allisonacs:
This looks pretty good to me @jzempel. It's not exact, but does a good impression. I don't know how much of this is under your control, but comparing these two:

**Sandbox**
image

**Garden**
image

The main differences I see are:
- CodeSandbox colors are more saturated
- Garden code doesn't match the `default` in `export default Example;` to e.g. `Button` in `import { Button }`

Neither of these are a big deal, and tbh the less saturated colors on our site are easier on ~the~ my eyes.

jzempel:
@allisonacs I'm using the "Dark +" (default) color theme in CodeSandbox. My diff appears to be less substantial:

**CodeSandbox**

Screen Shot 2020-07-15 at 11 33 46 AM

**Garden**

Screen Shot 2020-07-15 at 11 34 18 AM

The elements I attempted to pick off for styling but don't have access to with the parser we are using:

- blue `const` (i.e. declaration keyword tokens)
- blue `=>` arrow function symbol
- `<` & `/>` characters surrounding JSX
- before/after `"` are not included in attribute strings (i.e. `"center"`)

PR #122, opened 2020-07-22
chore: add domain-level redirects

jzempel:
## Description

301 redirect `zendeskgarden.netlify.app` -> `garden.zendesk.com`

### Details

- https://docs.netlify.com/routing/redirects/redirect-options/#domain-level-redirects
- https://www.gatsbyjs.org/packages/gatsby-plugin-netlify/#redirects

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -1,0 +1,2 @@
+# redirect default Netlify subdomain to primary domain
+https://zendeskgarden.netlify.app/* https://garden.zendesk.com/:splat 301!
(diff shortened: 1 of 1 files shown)


austin94:
@jzempel should this be handled by the `gatsby-plugin-netlify` plugin? https://www.gatsbyjs.org/packages/gatsby-plugin-netlify/

jzempel:
> should this be handled by the gatsby-plugin-netlify plugin

See from the PR description links, it is. It's just a simpler config option.

PR #139, opened 2020-08-04
chore: omit toggle best practice

rossmoody:
## Description

**Fixes Jira [GARDEN-1391]**
Took a look at the Toggle doc. We didn't outline any graphics for the point being made here and since we don't really advocate for the use of Toggles, thinking we ditch it for now until we can revisit with a more rounded takeaway.

## Checklist

### DIFF
--- a/src/pages/components/toggle.mdx
+++ b/src/pages/components/toggle.mdx
@@ -100,22 +100,6 @@ import ValidationCode from '!!raw-loader!../../examples/forms/toggle/Validation.
   <Validation />
 </CodeExample>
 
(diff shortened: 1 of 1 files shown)


allisonacs:
Is the removed content saved somewhere for the team to be able to access in the future? We may rather include this in a forthcoming pattern, instead of on the component docs. Let's make sure it's not lost to the sands of time.

rossmoody:
It's still hanging out in the Toggle Google doc for reference. There's quite a bit of punt activity in that doc too for consideration.

PR #142, opened 2020-08-04
chore: add menu deps to dropdown

rossmoody:
## Description

**Fix Jira [GARDEN]1377**
All "Dropdown" pages now include the same available Menu/Item exports 

## Checklist

### DIFF
--- a/src/pages/components/autocomplete.mdx
+++ b/src/pages/components/autocomplete.mdx
@@ -5,13 +5,24 @@ description: >
   something quickly in a large list of options.
 package: '@zendeskgarden/react-dropdowns'
 components:
+  - dropdowns/src/elements/Menu/Items/AddItem.tsx
   - dropdowns/src/elements/Autocomplete/Autocomplete.tsx
   - dropdowns/src/elements/Dropdown/Dropdown.tsx
   - dropdowns/src/elements/Fields/Field.tsx
+  - dropdowns/src/elements/Menu/Items/HeaderIcon.tsx
+  - dropdowns/src/elements/Menu/Items/HeaderItem.tsx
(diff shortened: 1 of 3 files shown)


rossmoody:
@jzempel I agree

austin94:
With `Autocomplete`, `Menu`, `Select`, and `Multiselect` now being on unique pages I see some improvement with this change. Before this change I'm not sure how someone going directly to `Autocomplete` for example would know that they are able to include menu items such as `ItemMeta`, `MediaItem`, etc.

To help simplify we could remove the non-Item prop-sheets though. i.e. The `Autocomplete` wouldn't need to see `Select` and `Trigger`

jzempel:
The only way the additions make sense to me is if we pull in corresponding examples that highlight expected `Menu` usage within the various components.

In other words, expecting the design/developer audience of users to pick up "I can use menu media items under a multiselect" through the subtlety of a long stack of imports and API prop sheets is not a high-quality Garden website experience. 

For now, I think the pages are best as-is.

allisonacs:
+1 @jzempel. While I can understand the concern, I think adding these prop sheets with no context ultimately is less helpful. If we're really concerned that people will come to an autocomplete and be confused or miss out on capabilities not explicitly listed, I'd recommend we have a sentence/paragraph under the API header saying something like,

> Autocomplete is a type of dropdown. Learn more about Menu and Item capabilities and props on the Menu page.

And link to the menu page. 

But, I also think we could leave it as is.

jzempel:
@rossmoody is your preference to explore adding Allison's suggested link or close this PR for now?

rossmoody:
My preference would be to close this one

PR #149, opened 2020-08-06
feat: add split button examples

rossmoody:
## Description

**Fixes Jira [GARDEN-1401]**
Adds default, danger and disabled examples to the Split Button component page using other Button page copy syntax as a template for the text. 

## Checklist

### DIFF
--- a/src/examples/button/SplitButtonDanger.tsx
+++ b/src/examples/button/SplitButtonDanger.tsx
@@ -11,7 +11,7 @@ import { SplitButton, Button, ChevronButton } from '@zendeskgarden/react-buttons
 import { Dropdown, Trigger, Menu, Item } from '@zendeskgarden/react-dropdowns';
 
 const Example = () => {
-  const [rotated, setRotated] = useState<boolean | undefined>();
+  const [rotated, setRotated] = useState<boolean>();
 
   return (
     <Row>
--- a/src/examples/button/SplitButtonDanger.tsx
+++ b/src/examples/button/SplitButtonDanger.tsx
@@ -1,0 +1,41 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useState } from 'react';
+import { Row, Col } from '@zendeskgarden/react-grid';
+import { SplitButton, Button, ChevronButton } from '@zendeskgarden/react-buttons';
+import { Dropdown, Trigger, Menu, Item } from '@zendeskgarden/react-dropdowns';
+
+const Example = () => {
+  const [rotated, setRotated] = useState<boolean | undefined>();
+
+  return (
+    <Row>
+      <Col textAlign="center">
+        <SplitButton>
+          <Button isDanger>Harvest</Button>
+          <Dropdown
+            onStateChange={options =>
+              Object.prototype.hasOwnProperty.call(options, 'isOpen') && setRotated(options.isOpen)
+            }
+          >
+            <Trigger>
+              <ChevronButton aria-label="other actions" isRotated={rotated} isDanger />
+            </Trigger>
+            <Menu placement="bottom-end">
+              <Item value="prune">Prune</Item>
+              <Item value="water">Water</Item>
+              <Item value="fertilize">Fertilize</Item>
+            </Menu>
+          </Dropdown>
+        </SplitButton>
+      </Col>
+    </Row>
+  );
+};
+
+export default Example;
--- a/src/examples/button/SplitButtonDisabled.tsx
+++ b/src/examples/button/SplitButtonDisabled.tsx
@@ -1,0 +1,48 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useState } from 'react';
+import { Row, Col } from '@zendeskgarden/react-grid';
+import { SplitButton, Button, ChevronButton } from '@zendeskgarden/react-buttons';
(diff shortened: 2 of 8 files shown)


src/pages/components/split-button.mdx
```diff
@@ -39,6 +39,40 @@ components:
...
+import SplitButtonDangerCode from '!!raw-loader!../../examples/button/SplitButtonDanger.tsx';
+
+<CodeExample code={SplitButtonDangerCode}>
+  <SplitButtonDanger />
+</CodeExample>
+
+### Disabled
+
+A disabled Split Button prevents user interaction. It doesn’t appear in the tab order,
+can’t receive focus, and may not read aloud by a screenreader.
```
allisonacs, line 67: Proving that the human brain can insert missing words in literally any scenario, it appears that when I rewrote the disabled copy, I omitted the word "be" in the last line. I'll open a PR fixing this.

src/pages/components/split-button.mdx
```diff
@@ -39,6 +39,40 @@ components:
...
+import SplitButtonDisabled from '../../examples/button/SplitButtonDisabled.tsx';
+import SplitButtonDisabledCode from '!!raw-loader!../../examples/button/SplitButtonDisabled.tsx';
+
+<CodeExample code={SplitButtonDisabledCode}>
+  <SplitButtonDisabled />
+</CodeExample>
+
 ### Size
 
 Split Buttons can be small, default, or large.
```
allisonacs, line 78: Maybe this is happening in a different PR somewhere else, but we updated these to use the word medium and say what the default is.

src/pages/components/split-button.mdx
allisonacs: Actually, I'm hitting all kinds of roadblocks to get my branch created, yarn started etc. Could you please do this for me? the sentence should read:

> A disabled Split Button prevents user interaction. It doesn’t appear in the tab order,
can’t receive focus, and may not **_be_** read aloud by a screenreader.

src/pages/components/split-button.mdx
```diff
@@ -39,6 +39,40 @@ components:
...
+Use danger styling for Split Buttons that enable destructive action.
+
+import SplitButtonDanger from '../../examples/button/SplitButtonDanger.tsx';
+import SplitButtonDangerCode from '!!raw-loader!../../examples/button/SplitButtonDanger.tsx';
+
+<CodeExample code={SplitButtonDangerCode}>
+  <SplitButtonDanger />
+</CodeExample>
+
+### Disabled
```
allisonacs, line 64: As I think a little more about this… it is possible to have one side and not the other disabled. So it is technically possible to have an enabled primary action, but a disabled chevron button, and vice versa. We even see this in product (I hate it). I don't know if we want to call out that it's possible and then also create a "how to use it well" section? Maybe that's a different PR?

src/examples/button/SplitButtonDanger.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useState } from 'react';
+import { Row, Col } from '@zendeskgarden/react-grid';
+import { SplitButton, Button, ChevronButton } from '@zendeskgarden/react-buttons';
+import { Dropdown, Trigger, Menu, Item } from '@zendeskgarden/react-dropdowns';
+
+const Example = () => {
+  const [rotated, setRotated] = useState<boolean | undefined>();
```
hzhu, line 14: What do you think about typing this out as only a `boolean`? It seems like initializing `useState` with no value (e.g. `undefined`) the TypeScript compiler use type inference to allow `undefined` values.

src/pages/components/split-button.mdx
rossmoody: Wow, that's nuts. I missed it too. Will fix

src/pages/components/split-button.mdx
rossmoody: Good spot. We did fix this in other places but this is new.

src/pages/components/split-button.mdx
rossmoody: My opinion on this one would be to not illustrate too much configuration for disabled usage and let the forthcoming "Disabled patterns" doc address these considerations for best practices more broadly.

src/examples/button/SplitButtonDanger.tsx
rossmoody: I followed an existing pattern for this convention. Git blame is telling me @jzempel set this up. I don't think I'm qualified to weigh in on this particular one.

jzempel: ...which would've been a mindless copy/paste motion on my part. Let's go with Henry's instinct on this one and clean up the code, if possible.

src/examples/button/SplitButtonDisabled.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Dropdown, Trigger, Menu, Item } from '@zendeskgarden/react-dropdowns';
+
+const Example = () => {
+  const [rotated, setRotated] = useState<boolean>();
+
+  return (
+    <Row>
+      <Col textAlign="center">
+        <SplitButton>
+          <Button disabled aria-disabled="true">
```
jzempel, line 20: Setting `disabled` [negates](https://codepen.io/stevef/post/short-note-on-the-disabled-attribute#also-note-5) any use of `aria-disabled`. The `aria-disabled` attribute may be removed here.

src/examples/button/SplitButtonDisabled.tsx
```diff
@@ -0,0 +1,48 @@
...
+            onStateChange={options =>
+              Object.prototype.hasOwnProperty.call(options, 'isOpen') && setRotated(options.isOpen)
+            }
+          >
+            <Trigger>
+              <ChevronButton
+                aria-label="other actions"
+                isRotated={rotated}
+                disabled
+                aria-disabled="true"
```
jzempel, line 33: Ditto above comment.

PR #154, opened 2020-08-14
feat: add multi-thumb range component

rossmoody:
## Description

This PR introduces the Multi-Thumb Range component to the website. 

## Checklist


src/examples/forms/multithumbrange/MultiThumbRangeValidation.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Row, Col } from '@zendeskgarden/react-grid';
+import { Field, Label, Message, MultiThumbRange } from '@zendeskgarden/react-forms';
+
+type validationTypes = 'success' | 'warning' | 'error';
+
+const messages: Record<validationTypes, string> = {
+  success: 'Growing all the time',
+  warning: 'Growing regularly',
+  error: 'Growing slowly'
+};
```
austin94, line 18: With this only being referenced once, you may be able to include a memoized value in the example. Something like:

```tsx
const validationMessage = useMemo(() => {
    if (validation === 'success') {
        return 'Success message';
    } else if (validation === 'warning') {
        return 'Warning message';
    }

    return 'Error message';
}, [validation]);

return (
    <Field>
        ....
        <Message validation={validation}>{validationMessage}</Message>
    </Field>
);
```

src/nav/components.yml
```diff
@@ -52,6 +52,8 @@
           title: Checkbox
         - id: /components/input
           title: Input
+        - id: /components/multithumbrange
+          title: Multi-Thumb Range
```
austin94, line 56: For hyphenated values do we normally `uppercase` the second half?

src/examples/forms/multithumbrange/MultiThumbRangeValidation.tsx
rossmoody: Wow, nice. Ty 🏆

src/nav/components.yml
rossmoody: Great call. Will update `range` PR link as well.

rossmoody: Oh wait, that's funny, I read something different the first time through which is that maybe this should be `multithumb-range` because the space previously has been a dash (i.e. `components/button-group`). 

I was following [W3s capitilization lead for this one](https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/multithumb-slider.html). What do you think? Is `Multi-thumb Range` our alt consideration?

jzempel: Would be nice if our https://garden.zendesk.com/content/punctuation#hyphens-and-dashes content had the answer. @vsrixyz?

src/pages/components/multi-thumb-range.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+### MultiThumbRange
+
+<PropSheet components={props.data.mdx.components} componentName="multithumbrange" />
+
+import { graphql } from 'gatsby';
+
+export const pageQuery = graphql`
+  query($fileAbsolutePath: String) {
+    ...SidebarPageFragment
```
jzempel, line 110: fwiw, a quick glance at the TOC looks like if we bring the `padding-right` down from 20px to 16px, we'll avoid the ellipsis truncation on `MultiThumbRange`. Although, I could be wrong and the overall padding adjustment may not be desirable.

src/examples/forms/multithumbrange/MultiThumbRangeDefault.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useState } from 'react';
+import { Row, Col } from '@zendeskgarden/react-grid';
+import { Field, Label, MultiThumbRange } from '@zendeskgarden/react-forms';
+
+const Example = () => {
+  const [minVal, setMinVal] = useState<number | undefined>();
+  const [maxVal, setMaxVal] = useState<number | undefined>();
```
hzhu, line 14: Since the default value of these states is `undefined`, the generic can be updated to `useState<number>();` to take advantage of type inference. TypeScript will infer that the default value is `undefined` so it doesn't need to be explicitly passed.

src/examples/forms/multithumbrange/MultiThumbRangeValidation.tsx
```diff
@@ -0,0 +1,55 @@
...
+    setMaxVal(maxValue);
+
+    if (minValue <= 20) {
+      setValidation('success');
+    } else if (minValue > 20 && maxValue > 70) {
+      setValidation('warning');
+    } else {
+      setValidation('error');
+    }
+  }, []);
```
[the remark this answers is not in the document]
rossmoody: You thinking we should update the content to correlate with the spacing? Or should we update the behavior to be a little simpler?

src/examples/forms/multithumbrange/MultiThumbRangeHint.tsx
```diff
@@ -0,0 +1,38 @@
...
+            There are between {minVal} and {maxVal} flowers available
+          </Hint>
+          <MultiThumbRange
+            minValue={minVal}
+            maxValue={maxVal}
+            onChange={({ minValue, maxValue }) => {
+              setMinVal(minValue);
+              setMaxVal(maxValue);
+            }}
+          />
```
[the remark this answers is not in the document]
rossmoody: I specified a min and max other than the default on this example, but if you were thinking something else, lemme know.

src/examples/forms/multithumbrange/MultiThumbRangeValidation.tsx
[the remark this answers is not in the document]
rossmoody: I updated the example to change message type a little easier for now.

src/examples/forms/multithumbrange/MultiThumbRangeHint.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <Row justifyContent="center">
+      <Col sm={5}>
+        <Field>
+          <Label>Flowers</Label>
+          <Hint>
+            There are between {minVal} and {maxVal} flowers available
+          </Hint>
+          <MultiThumbRange
+            min={10}
+            max={150}
```
jzempel, line 26: Ah, nvmd. I actually misread the API. I thought there was a way to specify a min for the upper thumb and a max for the lower – which prevents having them meet in the middle and having hint text that reads something like "There are between 61 and 61 flowers available". Let's just call it an API shortcoming and remove these for now.

src/examples/forms/multithumbrange/MultiThumbRangeValidation.tsx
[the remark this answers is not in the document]
jzempel: I was thinking something more like `maxValue - minValue`:

```js
  const onChange = useCallback(({ minValue, maxValue }) => {
    setMinVal(minValue);
    setMaxVal(maxValue);

    const result = maxValue - minValue;

    if (result >= 70) {
      setValidation('success');
    } else if (result >= 30) {
      setValidation('warning');
    } else {
      setValidation('error');
    }
  }, []);
```

Otherwise, I can't really make sense of what the two controls are doing without looking at the code. With :point_up: it starts to feel a little more intuitive.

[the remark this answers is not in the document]
rossmoody: Nice, ty

src/examples/forms/multithumbrange/MultiThumbRangeHint.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useState } from 'react';
+import { Row, Col } from '@zendeskgarden/react-grid';
+import { Field, Label, Hint, MultiThumbRange } from '@zendeskgarden/react-forms';
+
+const Example = () => {
+  const [minVal, setMinVal] = useState<number | undefined>(20);
+  const [maxVal, setMaxVal] = useState<number | undefined>(80);
```
jzempel, line 14: As @hzhu mentioned, I think you can take advantage of type inference here with `useState(20)` and `useState(80)`

hzhu: Whoops, I think `minValue` and `maxValue` from the multi-thumb range component's `onChange` returns these values typed as `number | undefined`. So I think typing the state out seems okay (otherwise it would need to be type casted when the state is yet).

src/nav/components.yml
vsrixyz: This should be `Multi-thumb range.` The first part of the expression is a single word and would follow our sentence case standard.

src/pages/components/multi-thumb-range.mdx
```diff
@@ -0,0 +1,112 @@
+---
+title: Multi-Thumb Range
+description: >
+  A Multi-Thumb Range is a slider with two or more thumbs that each set a value in
+  a group of related values.
```
vsrixyz, line 5: Replace with: A Multi-thumb Range lets users define a range between two values by using a slider.

src/pages/components/multi-thumb-range.mdx
```diff
@@ -0,0 +1,112 @@
...
+  - forms/src/elements/common/Hint.tsx
+  - forms/src/elements/common/Label.tsx
+  - forms/src/elements/common/Message.tsx
+  - forms/src/elements/MultiThumbRange.tsx
+---
+
+<Usage>
+  <Use>
+    <ul>
+      <li>To choose a range of numbers between minimum and maximum values</li>
```
vsrixyz, line 18: <li>To define a range of numbers between a minimum and maximum</li>

src/pages/components/multi-thumb-range.mdx
```diff
@@ -0,0 +1,112 @@
...
+    </ul>
+  </Use>
+  <Misuse>
+    <ul>
+      <li>
+        <Markdown>
+          To choose a single number between a range of values, use a [Range](/components/range)
```
vsrixyz, line 25: To choose a single number between two values, use a Range instead

src/pages/components/multi-thumb-range.mdx
```diff
@@ -0,0 +1,112 @@
...
+      </li>
+    </ul>
+  </Misuse>
+</Usage>
+
+## How to use it
+
+### Default
+
+The typical usage of a Multi-Thumb Range component.
```
vsrixyz, line 37: A Multi-thumb Range has two thumbs.

src/pages/components/multi-thumb-range.mdx
```diff
@@ -0,0 +1,112 @@
...
+import Disabled from '../../examples/forms/multithumbrange/MultiThumbRangeDisabled.tsx';
+import DisabledCode from '!!raw-loader!../../examples/forms/multithumbrange/MultiThumbRangeDisabled.tsx';
+
+<CodeExample code={DisabledCode}>
+  <Disabled />
+</CodeExample>
+
+### Hint text
+
+To provide additional context, use hint text with the label.
```
vsrixyz, line 60: Add context to the Multi-thumb Range label using Hint.

rossmoody:
@jzempel I lumped in a global sentence casing update into this PR to align.

PR #155, opened 2020-08-17
chore(docs): add TailwindCSS plugin link

austin94:
## Description

This PR adds the Garden Tailwind CSS plugin to the components sidebar.

## Detail

I've gone with `Tailwind CSS Plugin` for the naming since we already have `Bedrock CSS`. Open to any changes that you all think is best.

Originally went with "Design Tokens", but I'm not sure that's descriptive enough.

## Checklist

### DIFF
--- a/src/nav/components.yml
+++ b/src/nav/components.yml
@@ -6,6 +6,8 @@
   items:
     - id: https://zendeskgarden.github.io/css-components/bedrock/
       title: Bedrock CSS
+    - id: https://github.com/zendeskgarden/tailwindcss
+      title: Tailwind CSS Plugin
     - id: https://zendeskgarden.github.io/react-containers
       title: Containers
     - id: https://zendeskgarden.github.io/react-components/theming


src/nav/components.yml
```diff
@@ -6,6 +6,8 @@
   items:
     - id: https://zendeskgarden.github.io/css-components/bedrock/
       title: Bedrock CSS
+    - id: https://github.com/zendeskgarden/tailwindcss
+      title: Tailwind CSS Plugin
```
jzempel, line 10: Per content [rules](https://garden.zendesk.com/content/grammar#capitalization), please use sentence case: "Tailwind CSS plugin"

allisonacs:
I think we should go with `Design tokens`. This naming will have much better recognition for folks. Tailwind is our specific solution to this, just like React is our solution to components.

austin94:
I've updated this to `Design tokens`

rossmoody:
I'll throw in my two cents to consider here. To me, `Design Tokens` makes me think I'll see design tokens as in the variables or a page outlining Design Token usage/definitions, but isn't the Tailwind repo mainly a set of utility classes that applies design token styling? Like, we aren't defining or managing design tokens in the tailwindcss repo, right?

jzempel:
@rossmoody would "Utility CSS" land a bit better for you?

rossmoody:
For me it does but I'm still not convinced I fully understand. I think it's because I've been visiting design token documentation on other DS sites and the endpoint we would be communicating here seems different. [This page is](https://designsystem.digital.gov/design-tokens/#using-design-tokens) what I've been referencing. I've been referencing [Spectrum's](https://spectrum.adobe.com/page/design-tokens/) page too.

But, yeah maybe, `Utility classes` or `Utility CSS` or `CSS Utilities`

jzempel:
@rossmoody one way to think about it is that some day we'll want to have a page/section dedicated to talking about "Design tokens" on the website just like these properties you've referenced. But we're punting for now and simply linking off to the repo. With that in mind, it might be that "Design tokens" is a serviceable option for now.

austin94:
@jzempel @rossmoody is "Design tokens" good for now?

rossmoody:
I think I'm the oddball in this instance. I'll defer to y'all.

PR #157, opened 2020-08-19
feat: add inline loader component page

rossmoody:
## Description

This PR adds the Inline loader component page to the website. For the moment, you will need to access it from the Overview Page link as our root navigation automatically navigates to the old site. 

## Checklist

### DIFF
--- a/src/pages/components/inline.mdx
+++ b/src/pages/components/inline.mdx
@@ -18,7 +18,7 @@ components:
     <ul>
       <li>
         <Markdown>
-          To communicate an action is being processed, use the
+          To show a loading state, use the
           [Dots](https://zendeskgarden.github.io/react-components/loaders/#dots-1) loader instead
         </Markdown>
       </li>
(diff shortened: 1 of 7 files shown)


src/pages/components/inline.mdx
```diff
@@ -0,0 +1,82 @@
...
+import Default from '../../examples/loaders/inline/InlineDefault.tsx';
+import DefaultCode from '!!raw-loader!../../examples/loaders/inline/InlineDefault.tsx';
+
+<CodeExample code={DefaultCode}>
+  <Default />
+</CodeExample>
+
+### Color
+
+By default the color will be inherited from the parent element’s color. Can be
```
jzempel, line 44: Nice curly quote. Also a reminder that the `gatsby-remark-smartypants` plugin will pick up any that get missed 😉

src/pages/components/inline.mdx
```diff
@@ -0,0 +1,82 @@
...
+import Color from '../../examples/loaders/inline/InlineColor.tsx';
+import ColorCode from '!!raw-loader!../../examples/loaders/inline/InlineColor.tsx';
+
+<CodeExample code={ColorCode}>
+  <Color />
+</CodeExample>
+
+### Size
+
+Can be explicitly set. By default, the size will be inherited from the font-size
```
jzempel, line 56: Should `font-size` get backtick code treatment? Or would it be better to not use CSS-speak and just say "font size"?

src/pages/components/inline.mdx
```diff
@@ -0,0 +1,82 @@
...
+import DefaultCode from '!!raw-loader!../../examples/loaders/inline/InlineDefault.tsx';
+
+<CodeExample code={DefaultCode}>
+  <Default />
+</CodeExample>
+
+### Color
+
+By default the color will be inherited from the parent element’s color. Can be
+explicitly set.
```
allisonacs, line 45: - Add a comma after "By default"
- "Can be explicitly set" is a fragment. @vsrixyz can tell us if that's cool with UXCS or not. If we'd prefer to have a typical sentence structure, we could go with something like:

> Color can be explicitly set, but will be inherited from the parent element's color by default.

src/pages/components/inline.mdx
```diff
@@ -0,0 +1,82 @@
...
+import ColorCode from '!!raw-loader!../../examples/loaders/inline/InlineColor.tsx';
+
+<CodeExample code={ColorCode}>
+  <Color />
+</CodeExample>
+
+### Size
+
+Can be explicitly set. By default, the size will be inherited from the font-size
+of the parent element.
```
allisonacs, line 57: Same comment re: fragment as above.

src/pages/components/inline.mdx
jzempel: Btw, this does not appear to be true for the `Inline` loader: "...the size will be inherited from the font-size of the parent element."

rossmoody: Great spot. Ty

PR #166, opened 2020-08-31
feat(accordions): add stepper component

rossmoody:
## Description

This PR adds the Stepper component to the website. 

## Checklist

### DIFF
--- a/src/examples/accordions/stepper/StepperDefault.tsx
+++ b/src/examples/accordions/stepper/StepperDefault.tsx
@@ -1,0 +1,65 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { Stepper } from '@zendeskgarden/react-accordions';
+import { Button } from '@zendeskgarden/react-buttons';
+import { Row, Col } from '@zendeskgarden/react-grid';
+
+const StyledBtnRow = styled.div`
+  display: grid;
+  grid-auto-flow: column;
+  grid-gap: ${p => p.theme.space.sm};
+  justify-content: start;
+  margin: ${p => p.theme.space.sm} 0 0 0;
+`;
+
+const Example = () => {
+  const [step, setStep] = useState(0);
+
+  return (
+    <Row>
+      <Col>
+        <Stepper activeIndex={step}>
+          <Stepper.Step key={'step-1'}>
+            <Stepper.Label>Make good use of your location</Stepper.Label>
+            <Stepper.Content>
+              The success of your garden depends greatly on location.
+              <StyledBtnRow>
+                <Button onClick={() => setStep(1)}>Next</Button>
+              </StyledBtnRow>
+            </Stepper.Content>
+          </Stepper.Step>
+          <Stepper.Step key={'step-2'}>
+            <Stepper.Label>Plan your Garden layout</Stepper.Label>
+            <Stepper.Content>
+              After choosing a site for you garden, the next step is to imagine how the arrangement
+              of crops will look in the garden.
+              <StyledBtnRow>
+                <Button onClick={() => setStep(0)}>Back</Button>
+                <Button onClick={() => setStep(2)}>Next</Button>
+              </StyledBtnRow>
+            </Stepper.Content>
+          </Stepper.Step>
+          <Stepper.Step key={'step-3'}>
+            <Stepper.Label>Buy great seeds</Stepper.Label>
+            <Stepper.Content>
+              Buy clean, hearty, disease-free seeds. Most seed from reliable seed companies will
+              meet these specifications.
+              <StyledBtnRow>
+                <Button onClick={() => setStep(1)}>Back</Button>
+              </StyledBtnRow>
+            </Stepper.Content>
+          </Stepper.Step>
+        </Stepper>
+      </Col>
+    </Row>
+  );
+};
+
+export default Example;
--- a/src/examples/accordions/stepper/StepperDefault.tsx
+++ b/src/examples/accordions/stepper/StepperDefault.tsx
@@ -12,6 +12,7 @@ import { Button } from '@zendeskgarden/react-buttons';
 
 const StyledButtons = styled.div`
   margin-top: ${p => p.theme.space.sm};
+  padding: ${p => p.theme.shadowWidths.md};
 
   & > button {
     margin-${p => (p.theme.rtl ? 'right' : 'left')}: ${p => p.theme.space.base * 3}px;
--- a/src/examples/accordions/stepper/StepperDefault.tsx
+++ b/src/examples/accordions/stepper/StepperDefault.tsx
@@ -26,18 +26,22 @@ const Example = () => {
       <Col>
         <Stepper activeIndex={step}>
           <Stepper.Step key="step-1">
-            <Stepper.Label>Make good use of your location</Stepper.Label>
+            <Stepper.Label>Choose a good location</Stepper.Label>
             <Stepper.Content>
-              <StyledContent>The success of your garden depends greatly on location.</StyledContent>
+              <StyledContent>
+                Your garden&apos;s success depends on its location, so choose a spot that has
+                healthy soil, gets good light, and is easily watered.
+              </StyledContent>
               <Button onClick={onNext}>Next</Button>
             </Stepper.Content>
           </Stepper.Step>
           <Stepper.Step key="step-2">
-            <Stepper.Label>Plan your Garden layout</Stepper.Label>
+            <Stepper.Label>Plan your garden&apos;s layout</Stepper.Label>
             <Stepper.Content>
               <StyledContent>
-                After choosing a site for your garden, the next step is to imagine how the
-                arrangement of crops will look in the garden.
+                The layout of your garden depends on its purpose. If you&apos;re planting flowers,
+                consider aesthetics like color and layout. If you&apos;re growing food, think about
+                harvest times and the kinds of pests that might be attracted to your crops.
               </StyledContent>
               <Button onClick={onBack} style={{ marginRight: '12px' }}>
                 Back
--- a/src/examples/accordions/stepper/StepperDefault.tsx
+++ b/src/examples/accordions/stepper/StepperDefault.tsx
@@ -11,16 +11,15 @@ import { Stepper } from '@zendeskgarden/react-accordions';
 import { Button } from '@zendeskgarden/react-buttons';
 import { Row, Col } from '@zendeskgarden/react-grid';
 
-const StyledBtnRow = styled.div`
-  display: grid;
-  grid-auto-flow: column;
-  grid-gap: ${p => p.theme.space.sm};
-  justify-content: start;
-  margin: ${p => p.theme.space.sm} 0 0 0;
+const StyledContent = styled.div`
+  margin: 0 0 ${p => p.theme.space.sm} 0;
 `;
 
 const Example = () => {
   const [step, setStep] = useState(0);
+
+  const onNext = () => setStep(step + 1);
+  const onBack = () => setStep(step - 1);
 
   return (
     <Row>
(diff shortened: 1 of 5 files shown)


src/examples/accordions/stepper/StepperDefault.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <Row>
+      <Col>
+        <Stepper activeIndex={step}>
+          <Stepper.Step key={'step-1'}>
+            <Stepper.Label>Make good use of your location</Stepper.Label>
+            <Stepper.Content>
+              The success of your garden depends greatly on location.
+              <StyledBtnRow>
```
hzhu, line 33: What's the purpose of this wrapping element? It seems to be like it should be removed. It's currently a wrapping div that applies CSS grid, flex, and margin styling. Doesn't seem like a single button needs CSS grid or flex styling. 

Maybe there could be an inline style or a `StyledButton` to apply margin if necessary.

src/examples/accordions/stepper/StepperDefault.tsx
```diff
@@ -0,0 +1,65 @@
...
+            <Stepper.Content>
+              After choosing a site for you garden, the next step is to imagine how the arrangement
+              of crops will look in the garden.
+              <StyledBtnRow>
+                <Button onClick={() => setStep(0)}>Back</Button>
+                <Button onClick={() => setStep(2)}>Next</Button>
+              </StyledBtnRow>
+            </Stepper.Content>
+          </Stepper.Step>
+          <Stepper.Step key={'step-3'}>
```
hzhu, line 49: Since the value is a string this attribute can be added as a string:

```jsx
<Stepper.Step key="step-3">
```

src/examples/accordions/stepper/StepperIcons.tsx
```diff
@@ -0,0 +1,31 @@
...
+  <Row justifyContent="center">
+    <Col sm={10} textAlign="center">
+      <Stepper isHorizontal>
+        <Stepper.Step key={'step-1'}>
+          <Stepper.Label icon={<LeafIcon />}>Make good use of your location</Stepper.Label>
+        </Stepper.Step>
+        <Stepper.Step key="step-2">
+          <Stepper.Label icon={<LeafIcon />}>Plan your Garden layout</Stepper.Label>
+        </Stepper.Step>
+        <Stepper.Step key={'step-3'}>
```
hzhu, line 23: Looks like since `"step-3"` is a string it can be written like it is for "step-2".

src/examples/accordions/stepper/StepperDefault.tsx
```diff
@@ -0,0 +1,65 @@
...
+            </Stepper.Content>
+          </Stepper.Step>
+          <Stepper.Step key={'step-2'}>
+            <Stepper.Label>Plan your Garden layout</Stepper.Label>
+            <Stepper.Content>
+              After choosing a site for you garden, the next step is to imagine how the arrangement
+              of crops will look in the garden.
+              <StyledBtnRow>
+                <Button onClick={() => setStep(0)}>Back</Button>
+                <Button onClick={() => setStep(2)}>Next</Button>
```
hzhu, line 45: Rather than manually setting indices, I think it could be simplified by defining the functions in scope and using them instead. 

```jsx
const onNext = () => setStep(step + 1); // used for all "Next" buttons
const onBack = () => setStep(step - 1); // used for all "Back" buttons
```

jzempel: Have a look at `CodeBlock`'s CSS-only solution to a similar problem: https://github.com/zendeskgarden/react-components/blob/d3918bfeca6a7c446e72152f13eeb6598ac8d2d9/packages/typography/src/styled/StyledCodeBlockLine.ts#L14-L28

src/examples/accordions/stepper/StepperLayout.tsx
```diff
@@ -0,0 +1,84 @@
...
+  display: grid;
+  grid-auto-flow: column;
+  grid-gap: ${p => p.theme.space.sm};
+  margin: ${p => p.theme.space.sm} 0;
+`;
+
+const Example = () => {
+  const [step, setStep] = useState(0);
+
+  const horizontalContent = (index: number) => {
```
hzhu, line 30: Is there a way this code for rendering content can be more declarative? The horizontal `Stepper` example code from the old website doesn't have `if else` checks.

src/examples/accordions/stepper/StepperDefault.tsx
rossmoody: Was thinking of having one styled wrapper for the buttons that could apply to each instead of inline styling throughout the example code or a few separate wrappers. If I inline style on the button with margin, [it doesn't break the line](https://share.getcloudapp.com/bLuRZEON) which looked weird to me.  Open to a different/preferred approach.

src/examples/accordions/stepper/StepperDefault.tsx
rossmoody: Nice!

src/examples/accordions/stepper/StepperDefault.tsx
hzhu: Ohh I see, in this case, the floating text should be wrapped in a block element. It might look something like:

```jsx
<div>The success of your garden depends greatly on location.</div>
<Button>Next</Button>
```

src/examples/accordions/stepper/StepperDefault.tsx
```diff
@@ -0,0 +1,64 @@
...
+              <Button onClick={onNext}>Next</Button>
+            </Stepper.Content>
+          </Stepper.Step>
+          <Stepper.Step key="step-3">
+            <Stepper.Label>Buy great seeds</Stepper.Label>
+            <Stepper.Content>
+              <StyledContent>
+                Buy clean, hearty, disease-free seeds. Most seed from reliable seed companies will
+                meet these specifications.
+              </StyledContent>
```
allisonacs, line 54: Nice, this works!

allisonacs (changes requested):
Minor copy changes to the examples, and check with Ginny re: buttons.

src/examples/accordions/stepper/StepperDefault.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const Example = () => {
+  const [step, setStep] = useState(0);
+
+  const onNext = () => setStep(step + 1);
+  const onBack = () => setStep(step - 1);
+
+  return (
+    <Row>
+      <Col>
```
jzempel, line 26: Row-Col is not providing any value here; it can be removed.

src/examples/accordions/stepper/StepperDefault.tsx
```diff
@@ -0,0 +1,68 @@
...
+          </Stepper.Step>
+          <Stepper.Step key="step-2">
+            <Stepper.Label>Plan your garden&apos;s layout</Stepper.Label>
+            <Stepper.Content>
+              <StyledContent>
+                The layout of your garden depends on its purpose. If you&apos;re planting flowers,
+                consider aesthetics like color and layout. If you&apos;re growing food, think about
+                harvest times and the kinds of pests that might be attracted to your crops.
+              </StyledContent>
+              <Button onClick={onBack} style={{ marginRight: '12px' }}>
```
jzempel, line 46: This breaks with RTL layouts. Recommend replacing entire structure of `<Stepper.Content>` with something like the following:

```jsx
<Paragraph>
  ...
</Paragraph>
<StyledButtons>
  <Button>...</Button>
</StyledButtons>
```

The `StyledButtons` is a div that applies `margin-top` and handles margining between `${Button}` children. Something like (probably needs fine-tuning):

```css
{
  margin-top: ${p => p.theme.space.sm};

  & > ${Button} {
    margin-${p => (p.theme.rtl ? 'right' : 'left')}: ${p => p.theme.space.base * 3}px;

    &:first-child {
      margin-${p => (p.theme.rtl ? 'right' : 'left')}: 0;
    }
  }
}
```

Similar to the construct we see here: https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/styled/StyledFooterItem.ts

src/examples/accordions/stepper/StepperIcons.tsx
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import React from 'react';
+import { Stepper } from '@zendeskgarden/react-accordions';
+import { ReactComponent as LeafIcon } from '@zendeskgarden/svg-icons/src/12/leaf-stroke.svg';
+import { Row, Col } from '@zendeskgarden/react-grid';
+
+const Example = () => (
+  <Row justifyContent="center">
+    <Col sm={10} textAlign="center">
```
jzempel, line 15: This `textAlign` prop can be removed.

jzempel (changes requested):
- Swap out `&apos;` HTML entity throughout.
- Recommend replacing all `Stepper.Content` with commented construct.

src/examples/accordions/stepper/StepperDefault.tsx
```diff
@@ -0,0 +1,64 @@
...
+            <Stepper.Label>Plan your Garden layout</Stepper.Label>
+            <Stepper.Content>
+              <StyledContent>
+                After choosing a site for your garden, the next step is to imagine how the
+                arrangement of crops will look in the garden.
+              </StyledContent>
+              <Button onClick={onBack} style={{ marginRight: '12px' }}>
+                Back
+              </Button>
+              <Button onClick={onNext}>Next</Button>
```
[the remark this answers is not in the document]
ginnywood: We recommend a bottom action bar for Steppers. I’ll send you the pattern documentation.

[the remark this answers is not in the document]
rossmoody: Are we thinking we should add a Chrome Footer component to the example?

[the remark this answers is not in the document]
jzempel: I'd have to see the pattern docs, but I doubt it. The dimensions on the Chrome footer are quite large compared to the space we're dealing with here.

src/examples/accordions/stepper/StepperDefault.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <Col>
+        <Stepper activeIndex={step}>
+          <Stepper.Step key="step-1">
+            <Stepper.Label>Choose a good location</Stepper.Label>
+            <Stepper.Content>
+              <StyledContent>
+                Your garden&apos;s success depends on its location, so choose a spot that has
+                healthy soil, gets good light, and is easily watered.
+              </StyledContent>
+              <Button onClick={onNext}>Next</Button>
```
hzhu, line 35: Seems like there's something wrong with the focusvisible focus state in CodeSandbox. Also wasn't able to navigate to the button using the keyboard on the website 🤔 .

Screen Shot 2020-09-03 at 4 24 41 PM

src/examples/accordions/stepper/StepperLayout.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <Stepper.Label>Plan your garden&apos;s layout</Stepper.Label>
+          </Stepper.Step>
+          <Stepper.Step key="step-3">
+            <Stepper.Label>Buy great seeds</Stepper.Label>
+          </Stepper.Step>
+        </Stepper>
+        {allSteps.map(
+          (step, index) =>
+            index === currentStep && (
+              <StyledContainer>
```
hzhu, line 70: These mapped elements will need keys to help React identify which items have changed, are added, or are removed. The warnings in the console should go away once a unique key is added.

src/examples/accordions/stepper/StepperDefault.tsx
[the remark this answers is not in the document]
allisonacs: Ah cool. Since we don't want to bring in Chrome here, what you've got for demo purposes is 👌

src/examples/accordions/stepper/StepperDefault.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { Stepper } from '@zendeskgarden/react-accordions';
+import { Button } from '@zendeskgarden/react-buttons';
+
+const StyledButtons = styled.div`
+  margin-top: ${p => p.theme.space.sm};
+
+  & > button {
```
jzempel, line 16: Ideally, this would be `& > ${Button}`

src/examples/accordions/stepper/StepperLayout.tsx
```diff
@@ -0,0 +1,88 @@
...
+import styled from 'styled-components';
+import { Stepper } from '@zendeskgarden/react-accordions';
+import { Button } from '@zendeskgarden/react-buttons';
+import { Row, Col } from '@zendeskgarden/react-grid';
+
+const StyledButtons = styled.div`
+  margin-top: ${p => p.theme.space.sm};
+  padding: 3px;
+
+  & > button {
```
jzempel, line 18: Can this be `& > ${Button}`?

src/examples/accordions/stepper/StepperLayout.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { Stepper } from '@zendeskgarden/react-accordions';
+import { Button } from '@zendeskgarden/react-buttons';
+import { Row, Col } from '@zendeskgarden/react-grid';
+
+const StyledButtons = styled.div`
+  margin-top: ${p => p.theme.space.sm};
+  padding: 3px;
```
jzempel, line 16: This can be `padding: ${p => p.theme.shadowWidths.md};`

src/examples/accordions/stepper/StepperDefault.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { Stepper } from '@zendeskgarden/react-accordions';
+import { Button } from '@zendeskgarden/react-buttons';
+
+const StyledButtons = styled.div`
+  margin-top: ${p => p.theme.space.sm};
```
jzempel, line 14: Looks like https://github.com/zendeskgarden/website/pull/166#discussion_r483304391 is still a problem.

jzempel (changes requested):

src/examples/accordions/stepper/StepperDefault.tsx
rossmoody: Good spot, missed adding the padding here.

src/examples/accordions/stepper/StepperLayout.tsx
rossmoody: When I try to target it that way, my build doesn't succeed. It doesn't outright fail in my terminal, but in the browser it doesn't complete. This is the error message:

```
Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component.
```

jzempel: wow. Ok, good for now. Thx.

jzempel (approved):

vsrixyz (approved):

hzhu (approved):

src/pages/components/stepper.mdx
```diff
@@ -0,0 +1,75 @@
...
+  A Stepper guides users through steps of a task in sequential order.
+package: '@zendeskgarden/react-accordions'
+components:
+  - accordions/src/elements/stepper/Stepper.tsx
+---
+
+<Usage>
+  <Use>
+    <ul>
+      <li>To guide users through a task using progressive disclosure </li>
```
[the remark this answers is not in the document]
jzempel: @rossmoody I'll plan to merge after this final string replacement.

PR #167, opened 2020-08-31
feat(grid): add grid component

rossmoody:
## Description

This PR introduces the Grid component to the website. 

## Checklist

### DIFF
--- a/src/examples/grid/GridAllBreak.tsx
+++ b/src/examples/grid/GridAllBreak.tsx
@@ -1,0 +1,39 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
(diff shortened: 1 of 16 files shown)


rossmoody:
This is a pretty complex page with a lot of examples that are mirroring Bootstrap examples. I opted to go for a straight transfer of this page for now instead of pausing to do such a mammoth rewrite/new approach. Open to any thoughts or updates.

jzempel (approved):
Echoing @allisonacs's concerns. I think it's a bit of a mistake that we are connecting so hard to bootstrap. It begs the question "why should I use this Garden grid?" I think we all understand this thing is much more powerful. For example, the page doesn't highlight key features such as the ability to vary number of columns and gutter width. In addition, it would be helpful if the examples mapped more closely to typical layouts we observe in-product. But I understand that getting to that goal represents a lot more work.

PR #172, opened 2020-09-15
feat(table): add table component

rossmoody:
## Description

This PR introduces the Table component page to the website. 

## Checklist


rossmoody:
This is quite a component page so I was hoping to get a look at this initial approach on the first example to flush out any problems before continuing. Please let me know how this is looking and anything I can update and then I'll keep adding examples.

src/examples/table/TableDefault.tsx
```diff
@@ -0,0 +1,45 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Row as GRow, Col } from '@zendeskgarden/react-grid';
+import { Body, Cell, Head, HeaderCell, HeaderRow, Row, Table } from '@zendeskgarden/react-tables';
+
+const Example = () => (
+  <GRow style={{ overflow: 'auto' }}>
+    <Col style={{ minWidth: '500px' }}>
```
jzempel, line 14: No need for a Grid here. This can probably just be `<div style={{ minWidth: 500 }}>`. If there are overflow scrolling issues, we should fix that at the `CodeExample` component level.

jzempel: ...or maybe solve with `<div style={{ minWidth: 500, overflow: 'auto' }}>`. I haven't pulled to try locally.

rossmoody: Nice that seems much more neat. How's the updated looking?

jzempel: Nice. Looks like we have a to-edge scrolling issue. But that's not your example problem. We need to update the `CodeExample` component to handle more gracefully.

[nit] I would use `style={{ minWidth: 500 }}`, but that's just me :wink:

allisonacs (changes requested):
On the examples, I think it would be good to include at least 5 rows of data, if not more (10?), depending on what the example is demonstrating. For example, group rows and stripes should probably have at least 10 so that the user can more clearly see the features. But even the standard example feels scant—tables more typically display lots of data.

src/pages/components/table.mdx
```diff
@@ -0,0 +1,107 @@
+---
+title: Table
+description: >
+  Tables organize and display related content from a data set. This helps users
```
vsrixyz, line 4: _Edited for clarity:_

A Table organizes data into columns and rows. Tables make it easier for users to compare and analyze information.

src/pages/components/table.mdx
```diff
@@ -0,0 +1,107 @@
...
+  - tables/src/elements/OverflowButton.tsx
+  - tables/src/elements/Row.tsx
+  - tables/src/elements/SortableCell.tsx
+  - tables/src/elements/Table.tsx
+---
+
+<Usage>
+  <Use>
+    <ul>
+      <li>To organize and display data that has shared relationships</li>
```
vsrixyz, line 24: To organize and display data with logical relationships

vsrixyz (approved):
Small content tweaks.

rossmoody:
@allisonacs @jzempel @austin94 This is ready for a review. After meeting with Austin, I pulled the `Drag and drop` and `Virtual scrolling` examples out of this PR in favor of a follow-on in the next sprint where Eng can take a more comprehensive look at the example code.

src/pages/components/table.mdx
```diff
@@ -0,0 +1,210 @@
...
+import Scroll from '../../examples/table/TableScroll.tsx';
+import ScrollCode from '!!raw-loader!../../examples/table/TableScroll.tsx';
+
+<CodeExample code={ScrollCode}>
+  <Scroll />
+</CodeExample>
+
+### Selection
+
+Individual rows within a table can be selectable for bulk actions.
```
hzhu, line 99: Would "selected" be a suitable replacement for "selectable"?

@vsrixyz

src/examples/table/TableStripedRows.tsx
```diff
@@ -0,0 +1,52 @@
...
+const rowData: IRowData[] = [];
+
+for (let x = 0; x < 10; x++) {
+  rowData.push({
+    index: x,
+    fruit: 'Strawberries',
+    sun: 'Full sun',
+    soil: 'Well draining'
+  });
+}
```
hzhu, line 27: Similar to the comment above regarding imperative vs declarative, I think an `[].from()` with `[].map()` can be used to improve the example code.

src/examples/table/TablePagination.tsx
```diff
@@ -0,0 +1,74 @@
...
+    soil: 'Well draining'
+  });
+}
+
+const getPagedData = (data: any, currentPage: number, pageSize: number) => {
+  const output = [];
+
+  for (let x = (currentPage - 1) * pageSize; x < data.length && x < currentPage * pageSize; x++) {
+    output.push(data[x]);
+  }
```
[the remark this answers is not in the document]
hzhu: @austin94 I'm curious what you think too.

src/examples/table/TableSort.tsx
```diff
@@ -0,0 +1,137 @@
...
+const rowData: IRowData[] = [];
+
+for (let x = 0; x < 10; x++) {
+  rowData.push({
+    id: `row-${x}`,
+    subject: `Custom ticket view ${x + 1}`,
+    requester: x % 2 === 0 ? 'John Smith' : 'Jane Doe',
+    type: x % 3 === 0 ? 'Ticket' : 'Incident'
+  });
+}
```
hzhu, line 38: I think we should update this with a more declarative style

```tsx
const rowData: IRowData[] = Array.from(Array(10)).map((row, index) => ({
  id: `row-${index}`,
  subject: `Custom ticket view ${index + 1}`,
  requester: index % 2 === 0 ? 'John Smith' : 'Jane Doe',
  type: index % 3 === 0 ? 'Ticket' : 'Incident'
}));
```

src/examples/table/TableSort.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  if (requesterSort) {
+    field = 'requester';
+    sortValue = requesterSort;
+  } else {
+    field = 'type';
+    sortValue = typeSort;
+  }
+
+  return tableData.sort((a: any, b: any) => {
```
hzhu, line 56: These `any` types should be removed because it wipes out all type information within the sort logic. These `any` types can be removed if `field` was typed like this:

```tsx
let field: 'requester' | 'type';
```

src/examples/table/TableSort.tsx
```diff
@@ -0,0 +1,137 @@
...
+                  setRequesterSort('desc');
+                  setTypeSort(undefined);
+                } else if (requesterSort === 'desc') {
+                  setData(data);
+                  setRequesterSort(undefined);
+                  setTypeSort(undefined);
+                } else {
+                  setData(data);
+                  setRequesterSort('asc');
+                  setTypeSort(undefined);
```
hzhu, line 94: I know these were mostly copied from the existing examples, but I noticed that some of these `setStates` are called in every possible logic branch. Which means the calls can be combined into a single call without affecting the logic (I think):

```tsx
<SortableCell
  onClick={() => {
    if (requesterSort === 'asc') {
      setRequesterSort('desc');
    } else if (requesterSort === 'desc') {
      setRequesterSort(undefined);
    } else {
      setRequesterSort('asc');
    }
    setTypeSort(undefined);
    setData(data);
  }}
  sort={requesterSort}
>
  Requester
</SortableCell>
```

src/examples/table/TablePagination.tsx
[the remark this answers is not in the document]
rossmoody: Made this update. Please let me know any more optimizations can be made.

src/examples/table/TableSort.tsx
rossmoody: @austin94 Lemme know what you think for this and if we should revert to the [single setState call](https://github.com/zendeskgarden/website/pull/172/commits/87d755f14a986714af40f9ddb6d1ff42cedbd366#diff-bc140959802676c8fdb6126b94a27e56R71). You mentioned sometimes it's a best practice to separate them out.

src/examples/table/TableSelectRows.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+    return accumulator;
+  }, 0);
+
+  return numSelectedRows > 0 && numSelectedRows < rows.length;
+};
+
+const isSelectAllChecked = (rows: any) => {
+  return rows.every((row: any) => row.selected);
+};
```
hzhu, line 47: I think this function can be typed out without using `any`:

```tsx
const isSelectAllChecked = (rows: IRowData[]) => rows.every(row => row.selected);
```

src/examples/table/TableSelectRows.tsx
```diff
@@ -0,0 +1,144 @@
...
+                <Checkbox
+                  indeterminate={isSelectAllIndeterminate(data)}
+                  checked={isSelectAllChecked(data)}
+                  onChange={e => {
+                    if (e.target.checked) {
+                      const updatedRows = data.map((row: any) => ({ ...row, selected: true }));
+
+                      setData(updatedRows);
+                    } else {
+                      const updatedRows = data.map((row: any) => ({ ...row, selected: false }));
```
hzhu, line 70: TypeScript uses type inference to determine the type for `row`. It seems safe to remove `any` type annotation.

src/components/MarkdownProvider/components/CodeExample/index.tsx
```diff
@@ -46,6 +46,7 @@ export const CodeExample: React.FC<{ code: string }> = ({ children, code }) => {
           css={css`
             padding: ${p => p.theme.space.md};
             direction: ${p => p.theme.rtl && 'rtl'};
+            /* overflow-x: auto; */
```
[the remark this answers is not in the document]
jzempel: for the full experience, try adding the `&::after` styling shown here: https://github.com/zendeskgarden/react-components/blob/bb1f1056ed6f346be2131ef319a71c12252884c6/packages/typography/src/styled/StyledCodeBlockLine.ts#L52-L56. This should allow scrolling into the padded area.

hzhu (commented):
Lookin' solid 👍

src/examples/table/TableGroupedRows.tsx
```diff
@@ -0,0 +1,92 @@
...
+    <Table>
+      <Head>
+        <HeaderRow>
+          <HeaderCell>Type</HeaderCell>
+          <HeaderCell>Sun exposure</HeaderCell>
+          <HeaderCell>Soil</HeaderCell>
+        </HeaderRow>
+      </Head>
+      <Body>
+        <GroupRow key="fruits">
```
allisonacs, line 31: Let's make these bold please. Regular is allowed when there are key value pairs like:

> **Fruit** Plums

If there is no secondary categorization within the group row, we should use bold type.

src/pages/components/table.mdx
```diff
@@ -0,0 +1,210 @@
...
+import Pagination from '../../examples/table/TablePagination.tsx';
+import PaginationCode from '!!raw-loader!../../examples/table/TablePagination.tsx';
+
+<CodeExample code={PaginationCode}>
+  <Pagination />
+</CodeExample>
+
+### Scroll
+
+To ensure that your table is read correctly by assistive technologies follow the
```
allisonacs, line 86: Add a comma:

> assistive technologies**,** follow…

src/pages/components/table.mdx
```diff
@@ -0,0 +1,210 @@
...
+import Scroll from '../../examples/table/TableScroll.tsx';
+import ScrollCode from '!!raw-loader!../../examples/table/TableScroll.tsx';
+
+<CodeExample code={ScrollCode}>
+  <Scroll />
+</CodeExample>
+
+### Selection
+
+Individual rows within a table can be selected for bulk actions.
```
allisonacs, line 99: The demo shows that you can also select all rows on a page with the checkbox in the header row. How would you feel about calling out this functionality in the docs?

src/examples/table/TablePagination.tsx
```diff
@@ -0,0 +1,63 @@
...
+  </Row>
+);
+
+const Example = () => {
+  const pageSize = 10;
+  const [currentPage, setCurrentPage] = useState(1);
+
+  return (
+    <div style={{ minWidth: 500 }}>
+      <Table style={{ marginBottom: '20px' }}>
```
jzempel, line 38: Please prefer using a `<StyledTable>` with a themed value for the bottom margin.

src/pages/components/table.mdx
```diff
@@ -0,0 +1,210 @@
...
+<CodeExample code={SizeCode}>
+  <Size />
+</CodeExample>
+
+### Sort
+
+Columns can be sorted by selecting table headers.
+
+import Sort from '../../examples/table/TableSort.tsx';
+import SortCode from '!!raw-loader!../../examples/table/TableSort.tsx';
```
allisonacs, line 124: There is some weirdness on the sorting controls. Recorded a screencast (with voice over) to demo the issue. https://share.getcloudapp.com/NQu1b1ll

I checked Styleguidist, and it looks like the focus thing is just how it works, but the second half of the video about the 2nd column's sorting functionality stands.

src/pages/components/table.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+<CodeExample code={SortCode}>
+  <Sort />
+</CodeExample>
+
+### Striped rows
+
+The background color of rows can alternate to help delineate content when
+scanning long rows. Striped and grouped rows shouldn’t be active at the same
+time.
```
allisonacs, line 134: Couple things here:
1. First sentence uses passive tense. Maybe instead: 

> You can alternate row background color to help make long rows easier to scan.

2. "shouldn't be active" feels a bit jargony… maybe change "active" to "used"

> Don't use striped and group rows at the same time.

src/examples/table/TableSort.tsx
```diff
@@ -0,0 +1,133 @@
...
+                  setRequesterSort(undefined);
+                  setTypeSort('desc');
+                } else if (typeSort === 'desc') {
+                  setData(data);
+                  setRequesterSort(undefined);
+                  setTypeSort(undefined);
+                } else {
+                  setData(data);
+                  setRequesterSort(undefined);
+                  setTypeSort('desc');
```
jzempel, line 110: Does this want to be `setTypeSort('asc')`?

src/pages/components/table.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+import StripedRows from '../../examples/table/TableStripedRows.tsx';
+import StripedRowsCode from '!!raw-loader!../../examples/table/TableStripedRows.tsx';
```
allisonacs, line 137: This is a lot of rows. Maybe cut it down to 10 or 20? I think 10 is probably enough to demonstrate the concept.

src/pages/components/table.mdx
```diff
@@ -0,0 +1,210 @@
...
+<CodeExample code={ScrollCode}>
+  <Scroll />
+</CodeExample>
+
+### Selection
+
+Individual rows within a table can be selected for bulk actions.
+
+import SelectRows from '../../examples/table/TableSelectRows.tsx';
+import SelectRowsCode from '!!raw-loader!../../examples/table/TableSelectRows.tsx';
```
allisonacs, line 102: Also realized that this example allows for multiple row selection. On the original set of docs, we say that this is additional functionality… should we do the same here?

src/examples/table/TableStripedRows.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Body, Cell, Head, HeaderCell, HeaderRow, Row, Table } from '@zendeskgarden/react-tables';
+
+interface IRow {
+  index: number;
+  fruit: string;
+  sun: string;
+  soil: string;
+}
+
+const rowData: IRow[] = Array.from(Array(100)).map((row, index) => ({
```
jzempel, line 18: Whoa, this example feels quite a bit too long.

jzempel (changes requested):

src/pages/components/table.mdx
```diff
@@ -0,0 +1,210 @@
...
+  - tables/src/elements/OverflowButton.tsx
+  - tables/src/elements/Row.tsx
+  - tables/src/elements/SortableCell.tsx
+  - tables/src/elements/Table.tsx
+---
+
+<Usage>
+  <Use>
+    <ul>
+      <li>To organize and display data with logical relationships</li>
```
allisonacs, line 24: I wonder if we should call out here that the data should be read-only?

allisonacs (changes requested):
Looks also like we are missing `Caption` and `Draggable` examples?

src/pages/components/table.mdx
jzempel: +1 I think this is worth drawing attention to if we can incorporate the copy nicely. Same could be said for calling attention to the `shift` key logic for selecting multiple rows.

src/pages/components/table.mdx
jzempel: The focus feels like a component bug. The action should match what we see under "Row Modifications" in this old CSS demo: https://zendeskgarden.github.io/css-components/tables/

rossmoody: I filed a ticket for this component bug in Jira. I'll be sure the sorting functionality is working correctly (I believe JZ addressed it in a comment in this thread).

allisonacs (approved):

src/examples/table/TableSort.tsx
hzhu: Consider this:

```js
if (hungry) {
  eat();
  relax();
} else if (thirsty) {
  drink();
  relax();
} else {
  relax();
};
```

There isn't a need to call `relax` on every logical branch if you know it must happen all the time. This would be preferred:

```js
if (hungry) {
  eat();
} else if (thirsty) {
  drink();
};

relax();
```

The same can be said of these function calls:

```tsx
setTypeSort(undefined);
setData(data);
```

hzhu (approved):

src/examples/table/TableSort.tsx
rossmoody: Nice! Much better

src/components/MarkdownProvider/components/CodeExample/index.tsx
```diff
@@ -46,6 +46,13 @@ export const CodeExample: React.FC<{ code: string }> = ({ children, code }) => {
           css={css`
             padding: ${p => p.theme.space.md};
             direction: ${p => p.theme.rtl && 'rtl'};
+            overflow: auto;
+
+            &::after {
+              display: inline-block;
+              width: ${props => props.theme.space.base * 3}px;
+              content: '';
+            }
```
jzempel, line 55: This doesn't seem to be working as suggested. Let's just remove for now and address later.

jzempel (approved):

PR #174, opened 2020-09-16
feat(modal): add modal component

rossmoody:
## Description

This PR adds the Modal component to the website. 

## Checklist

### DIFF
--- a/src/examples/modal/ModalDanger.tsx
+++ b/src/examples/modal/ModalDanger.tsx
@@ -16,7 +16,9 @@ const Example = () => {
   return (
     <Row>
       <Col textAlign="center">
-        <Button onClick={() => setVisible(true)}>Open modal</Button>
+        <Button onClick={() => setVisible(true)} isDanger>
+          Open danger modal
+        </Button>
         {visible && (
           <Modal onClose={() => setVisible(false)}>
             <Header isDanger>Remove plant food from your garden</Header>
--- a/src/examples/modal/ModalDanger.tsx
+++ b/src/examples/modal/ModalDanger.tsx
@@ -23,7 +23,8 @@ const Example = () => {
           <Modal onClose={() => setVisible(false)}>
             <Header isDanger>Remove plant food from your garden</Header>
             <Body>
-              Plant food is great for the health of a garden. Are you sure you want to remove it?
+              Plant food helps gardens grow. Removing plant food negatively affects the plant
+              health.
             </Body>
             <Footer>
               <FooterItem>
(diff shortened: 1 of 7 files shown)


jzempel:
Something borked with the deployment. Re-running now...

hzhu (commented):
It looks like there's an issue with the focusvisible behavior when the modal is opened. The focus outline isn't removed on blur.

Edit: after some investigation, it seems to be a [regression](https://zendeskgarden.github.io/react-components/modals/) unrelated to this PR. The root cause of the problem is TBD.

Edit 2: Looks like could be an issue with [react-merge-refs](https://github.com/zendeskgarden/react-components/pull/876/files#diff-b1dcdbfc5df6b70c458a6ffc64e4c2a6R154). A ref callback is being passed to `useFocusVisible` which means the focusvisible `scope` is undefined. I think it should take a ref object (rather than ref callback).

src/examples/modal/ModalDanger.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Modal, Header, Body, Footer, FooterItem, Close } from '@zendeskgarden/react-modals';
+import { Row, Col } from '@zendeskgarden/react-grid';
+
+const Example = () => {
+  const [visible, setVisible] = useState(false);
+
+  return (
+    <Row>
+      <Col textAlign="center">
+        <Button onClick={() => setVisible(true)}>Open modal</Button>
```
jzempel, line 19: Should this button be `isDanger` to differentiate from the other examples?

jzempel (approved):
:tada:

hzhu (approved):

rossmoody:
@vsrixyz @allisonacs This one was in limbo for a while due to build problems but it's ready for a look now.

src/examples/modal/ModalDanger.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <Row>
+      <Col textAlign="center">
+        <Button onClick={() => setVisible(true)} isDanger>
+          Open danger modal
+        </Button>
+        {visible && (
+          <Modal onClose={() => setVisible(false)}>
+            <Header isDanger>Remove plant food from your garden</Header>
+            <Body>
+              Plant food is great for the health of a garden. Are you sure you want to remove it?
```
allisonacs, line 26: We don't ask users if they are sure in Zendesk. Per our tone guidelines, we tell people what the effects of a destructive action are. Perhaps something more like

> Plant food helps gardens grow. Removing plant food negatively affects the plant health.

@vsrixyz what do you think?

allisonacs (changes requested):
Couple of content issues here, but implementation looks good.

src/examples/modal/ModalDanger.tsx
rossmoody: Made these changes for now to keep this moving.

allisonacs (approved):

PR #178, opened 2020-09-22
feat(forms): add auto-expanding text area

rossmoody:
## Description

This PR adds an example for auto-expanding textarea to the Forms documentation. 

## Checklist


src/examples/forms/inputs/AutoExpand.tsx
```diff
@@ -0,0 +1,23 @@
...
+import React from 'react';
+import { Row, Col } from '@zendeskgarden/react-grid';
+import { Field, Label, Textarea } from '@zendeskgarden/react-forms';
+
+const Example = () => (
+  <Row justifyContent="center">
+    <Col sm={5}>
+      <Field>
+        <Label>Ivy</Label>
+        <Textarea minRows={1} maxRows={12} />
```
austin94, line 17: Since textareas normally include multi-line text we set the `[minRows=3]` in the products. This shows that difference in input fields when there is no text included. 

Should we do that here as well?

src/pages/components/input.mdx
```diff
@@ -39,6 +39,18 @@ import InputCode from '!!raw-loader!../../examples/forms/inputs/Input.tsx';
   <Input />
 </CodeExample>
 
+### Auto-expanding textarea
```
austin94, line 42: The current example placement makes it seem like we provide an auto-expanding input. I would move this lower in position relative to the `Textarea input` section.

rossmoody: I battled with this and I'm glad you pointed it out. Does seem to make more sense (and more discoverable) within other textarea documentation.

src/pages/components/input.mdx
```diff
@@ -131,8 +131,9 @@ import SizeCode from '!!raw-loader!../../examples/forms/inputs/Size.tsx';
 
 ### Textarea input
 
-A Textarea is for multiline text. By default, its height is fixed and becomes resizable with the
-`isResizable` prop.
+A Textarea is for multiline text. Textarea height is fixed by default but can be
+configured to resize automatically based on minimum and maximum rows or resize
```
austin94, line 135: I'll lean on @vsrixyz more for this, but it may be cleaner to remove the mention of `minimum` and `maximum` rows. The height is determined by input, those values only define the bounds.

Something like:

> ... can be configured to resize automatically based on user input or manually with the `isResizable` prop

rossmoody: I very much prefer the direction of your new line. I'll make that update in preparation for UXCS feedback.

vsrixyz: I like it. Made a few tweaks:

A Textarea is for multiline text. By default, Textarea height is fixed but can be configured to resize automatically based on user input or manually with `isResizable`.

vsrixyz (approved):

austin94 (approved):

allisonacs (changes requested):
Per Slack convo, the content explaining how to invoke the resizeable vs. auto-expanding is not clear. Recommendation is to split out examples to default, resizeable, and auto-expanding so that we can be explicit about how to achieve each version.

allisonacs (changes requested):
Hey Ross, I'm sorry, looks like I wasn't clear enough in our convo. I think splitting these out into separate examples is the right thing to do, but not in the way you have it here (this makes it seem like there are separate components). I was thinking something more like:

# Textarea input
Explanation about default example
(shows default example)

Explanation about resizable text area
(shows resizeable example)

Explanation about auto-expanding
(shows auto-expand example)

I _think_ this will make it clear that it's one component with a few options, while allowing us to have clearer explanations (no need to keep repeating what's default first). I recognize it's a little out of the ordinary though. 

If this feels weird, LMK—we might have to choose between some awkward options. I wonder if Textarea should have its own page?

jzempel:
> I wonder if Textarea should have its own page?

Separate page would be my vote.

rossmoody:
@allisonacs @jzempel @vsrixyz Gave this one a more thorough refactor to separate out the Textarea component into its own page.

hzhu (approved):

jzempel (approved):
nice ❗

austin94 (approved):

allisonacs (approved):
So much clearer, thank you!

PR #188, opened 2020-10-01
feat(modals): add drawer component 

rossmoody:
## Description

This PR adds the Drawer component to the website. 

## Checklist


src/pages/components/drawer.mdx
```diff
@@ -0,0 +1,87 @@
...
+        To display information or actions that are supplementary to the screen’s primary content
+      </li>
+      <li>To display a list of actions that affect the screen’s content, such as filtering data</li>
+    </ul>
+  </Use>
+  <Misuse>
+    <ul>
+      <li>
+        <Markdown>
+          To block access to the primary view and force a decision from the user, use a
```
hzhu, line 23: I feel a bit unsure about this because a Drawer is actually a modal. Its component name is `DrawerModal` and it lives in the `react-modals` package. It uses the [core modal logic](https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/elements/DrawerModal/DrawerModal.tsx#L101) and is based on the [modal dialog](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal) pattern.

hzhu (commented):
When the drawer opens, the user is scrolled up to the top 🤔 .  It feels like this should not happen. I'm going to look into what is causing this.

allisonacs:
> When the drawer opens, the user is scrolled up to the top 🤔 . It feels like this should not happen. I'm going to look into what is causing this.

Yea, that's definitely not good.

allisonacs (changes requested):
Per conversation IRL, we've got some work to do to get this where it needs to be. We'll circle back once component changes are made.

src/pages/components/drawer.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+### Default
+
+When the Drawer opens, focus moves to it so users who rely on a keyboard and/or
+screen reader can access the content within. Focus is kept in the Drawer until
+it is dismissed.
+
+Users may be able to see the screen’s content when the Drawer is open but they
+will only be able to interact with the Drawer or the page content at any one
+time.
```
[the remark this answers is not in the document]
rossmoody: Fixed with latest transparency style alignment.

jzempel (approved):
All existing component issues will be alleviated with the upcoming `react-components` release.

rossmoody:
@allisonacs The build succeeded! Quick note, the background of the Drawer page scrolls. I reached out to Henry for the reason and it looks like it's a bug in the CSS bedrock that will need addressed separately. It's happening on the Modals page as well.

src/pages/components/drawer.mdx
```diff
@@ -0,0 +1,77 @@
...
+      </li>
+      <li>To display a list of actions that affect the screen’s content, such as filtering data</li>
+    </ul>
+  </Use>
+  <Misuse>
+    <ul>
+      <li>
+        <Markdown>
+          To obscure the main content and focus the user on modal content, use a
+          [Modal](/components/modal)
```
allisonacs, line 24: Does this need to be rewritten in light of the fact we are not obscuring main page content with a background? It is visible, but many users won't be able to consume it (low vision, but for sure screenreaders, which are focus jailed).

src/pages/components/drawer.mdx
```diff
@@ -0,0 +1,77 @@
...
+        </Markdown>
+      </li>
+      <li>
+        <Markdown>
+          To display a small amount of content that contains actions, use a Tooltip Modal
+        </Markdown>
```
allisonacs, line 30: This sentence doesn't flow well, I had to re-read a few times to understand. Can you reword it?

allisonacs (changes requested):
Thanks for the heads up on the bug, no sweat there. Two nits on the "not for this" section, otherwise gtg.

src/pages/components/drawer.mdx
rossmoody: Good point, to keep this moving I'll omit it for now and add it to the Google doc to refactor.

src/pages/components/drawer.mdx
rossmoody: I will omit this for now as well. I think this reaches a little too far for the utility of this section. The big benefit of `When not to use` section was to help make a decision when applications are close and I don't think that is a likely confusion here since Tooltips are for contextual use on the page.

allisonacs (approved):

PR #190, opened 2020-10-07
fix: amend component overflow style

rossmoody:
## Description

The `overflow: auto` styling added to the component example wrapper in the latest Table PR brought with it some undesirable behavior on other examples. This PR addresses that problems and handles the overflow behavior for the Tables in the Table example code instead. 

## Checklist

### DIFF
--- a/src/examples/table/TableCaption.tsx
+++ b/src/examples/table/TableCaption.tsx
@@ -19,7 +19,7 @@ import {
 } from '@zendeskgarden/react-tables';
 
 const Example = () => (
-  <div style={{ overflow: 'auto' }}>
+  <div style={{ overflowX: 'auto' }}>
     <Table style={{ minWidth: 500 }}>
       <Caption>
         <XL>Garden details</XL>
--- a/src/examples/table/TableCaption.tsx
+++ b/src/examples/table/TableCaption.tsx
@@ -19,8 +19,8 @@ import {
 } from '@zendeskgarden/react-tables';
 
 const Example = () => (
-  <div style={{ minWidth: 500 }}>
-    <Table>
+  <div style={{ overflow: 'auto' }}>
+    <Table style={{ minWidth: 500 }}>
       <Caption>
         <XL>Garden details</XL>
       </Caption>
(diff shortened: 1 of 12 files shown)


rossmoody:
@jzempel @austingreendev I'm unsure why this build is failing but I think it has something to do with timing out during build time trying to get assets. Is there anything I can do to get a build up for review?

austingreendev (approved):

austingreendev:
I'm not sure this is related to our asset generation. It looks like the Node process is running out of memory during JS compilation.

jzempel:
@rossmoody can you point out the existing overflow problems with a URL? We may want to finesse this a bit differently, depending on current issues.

rossmoody:
It looks like the [Split button](https://garden.zendesk.com/components/split-button) examples are overflowing inside and then Austin pointed out the [Dropdown example menus](https://garden.zendesk.com/components/autocomplete) are behaving/formatting oddly.

jzempel:
Would switching to `overflow-x` (the way you originally had it 😬) fix the container without harming menus? Feels like it would be ideal if we can fix the example container rather than chasing our tail per example.

rossmoody:
It would seem `overflow-x` still causes the Menus in the Dropdown [examples to display oddly](https://share.getcloudapp.com/qGuvoGge).

src/examples/table/TableCaption.tsx
```diff
@@ -19,8 +19,8 @@ import {
 } from '@zendeskgarden/react-tables';
 
 const Example = () => (
-  <div style={{ minWidth: 500 }}>
-    <Table>
+  <div style={{ overflow: 'auto' }}>
```
jzempel, line 22: [nit] these could probably all be `overflowX` as that is the only dimension we care about controlling 🤷

jzempel (approved):

src/examples/table/TableCaption.tsx
rossmoody: Nice, will amend right meow.

PR #191, opened 2020-10-07
feat(buttons): add ToggleButton and ToggleIconButton component pages

jzempel:
## Description

New pages added under the "Buttons" component category.

## Detail

@ginnywood I'm opening this PR to give you a couple of items to react to so we can get them addressed and properly documented. One is the use of `star` icon throughout the examples. Please lmk if you want that changed. The others I will highlight with inline code comments.

## Checklist

- [ ] :zap: ~audited via [web.dev](https://web.dev/measure/) for performance and SEO~


src/examples/button/ToggleIconButtonIcons.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const [pressed, setPressed] = useState(false);
+
+  return (
+    <Row>
+      <Col textAlign="center">
+        <ToggleIconButton isPressed={pressed} onClick={() => setPressed(!pressed)}>
+          <svg aria-label="leaf" focusable="false">
+            <FillIcon style={{ opacity: pressed ? 1 : 0 }} />
+            <StrokeIcon style={{ opacity: pressed ? 0 : 1 }} />
+          </svg>
```
jzempel, line 24: @ginnywood this example will receive the animation we settled on once `react-components` is updated to the latest version 8.24.1.

src/pages/components/toggle-icon-button.mdx
```diff
@@ -0,0 +1,151 @@
...
+### Icon
+
+A Toggle Icon Button may be structured to switch icons when pressed.
+
+import ToggleIconButtonIcons from '../../examples/button/ToggleIconButtonIcons.tsx';
+import ToggleIconButtonIconsCode from '!!raw-loader!../../examples/button/ToggleIconButtonIcons.tsx';
+
+<CodeExample code={ToggleIconButtonIconsCode}>
+  <ToggleIconButtonIcons />
+</CodeExample>
```
jzempel, line 71: @ginnywood I went rogue and added this section without content review, etc. Will wait for you to get the heading and description done just right. :pray:

allisonacs:
In terms of written content, these are both looking good to me. Couple nitpicks:
- the call out of "dark mode." While true, I'm not sure it makes as much sense as something "Bold," especially on the icon button page
- One of the "not for this" sections says that if you want to include an icon _and_ text, tiles should be used, but I don't think that's true. Tiles are form elements and act like Radio buttons, so they need a submit button, and once selected, you can't unselect (you can change which tile in a group you choose but you can't reset them). More intuitively, I would imagine that a toggle button could have a media item. Can we recommend that instead? 

I'm having a hard time still with the current ⭐️ icon being used as the example. Although intuitively I think it makes sense that Favorite would be an icon toggle button, I think the mental model is that this would be the icon that gets filled in. Since we don't want to imply that ⭐️ should be filled in, I think we should keep the leaf as the fill example, but perhaps we should switch the icon being used to the `B` for bold icon, or something else that you would more readily associate with a typical toggle button?

jzempel:
@ginnywood I'll hold for you to review Allison's comments: https://github.com/zendeskgarden/website/pull/191#issuecomment-705202411

src/pages/components/toggle-icon-button.mdx
ginnywood: I also went rogue and added a section to the Google Doc without content review. Mentioned you and Allison there.

ginnywood:
> * the call out of "dark mode." While true, I'm not sure it makes as much sense as something "Bold," especially on the icon button page
Love it. Updated copy in Google Doc
> * One of the "not for this" sections says that if you want to include an icon _and_ text, tiles should be used, but I don't think that's true. Tiles are form elements and act like Radio buttons, so they need a submit button, and once selected, you can't unselect (you can change which tile in a group you choose but you can't reset them). More intuitively, I would imagine that a toggle button could have a media item. Can we recommend that instead?
Good call! Updated Google Doc
> I'm having a hard time still with the current ⭐️ icon being used as the example. Although intuitively I think it makes sense that Favorite would be an icon toggle button, I think the mental model is that this would be the icon that gets filled in. Since we don't want to imply that ⭐️ should be filled in, I think we should keep the leaf as the fill example, but perhaps we should switch the icon being used to the `B` for bold icon, or something else that you would more readily associate with a typical toggle button?
Bold sounds good to me. That makes sense.

src/pages/components/toggle-icon-button.mdx
vsrixyz: Just looked over the content now and it looks good to me!

hzhu (approved):

allisonacs (approved):

ginnywood (approved):

PR #193, opened 2020-10-09
fix: allow overflow in code example

hzhu:
## Description

This PR proposes to allow overflow for code examples which would allow `SplitButton` to overflow rather than to be clipped.

## Detail

It looks like originally `overflow: hidden;` was applied to `CodeExample` to prevent code block from overflowing rounded container borders. To address both the `SplitButton` and the original rounded border overflow issue, I've removed `overflow: hidden;` and applied rounded borders to the `CodeBlock` component.

## Screenshots

**Before:** overflowed menus were clipped

--------

**After:** overflowed menus are no longer clipped

## Checklist

### DIFF
--- a/src/components/MarkdownProvider/components/CodeExample/index.tsx
+++ b/src/components/MarkdownProvider/components/CodeExample/index.tsx
@@ -6,6 +6,7 @@
  */
 
 import React, { useRef, useState, useMemo } from 'react';
+import { math } from 'polished';
 import styled, { css, DefaultTheme } from 'styled-components';
 import { ThemeProvider, DEFAULT_THEME, getColor } from '@zendeskgarden/react-theming';
 import { Tooltip } from '@zendeskgarden/react-tooltips';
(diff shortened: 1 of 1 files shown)


src/components/MarkdownProvider/components/CodeExample/index.tsx
```diff
@@ -18,6 +18,11 @@ import { ReactComponent as CodeSandboxIcon } from './assets/codesandbox-icon.svg
 import { retrieveCodesandboxParameters } from './utils/retrieveCodesandboxParameters';
 import { copyToClipboard } from './utils/copyToClipboard';
 
+const StyledCodeBlock = styled(CodeBlock)`
+  border-bottom-left-radius: ${props => props.theme.space.base}px;
+  border-bottom-right-radius: ${props => props.theme.space.base}px;
```
jzempel, line 23: These should be based upon `theme.borderRadii.md`. Also, when employing a technique like this, subtract 1 pixel in order to get the right fit. Otherwise, you end up with whitespace between the edges, which can be observed here on zoom.

jzempel (changes requested):

src/components/MarkdownProvider/components/CodeExample/index.tsx
hzhu: Oo thanks for the tip. The style now uses `theme.borderRadii.md` and gets rid of the whitespace.

jzempel: Weird. Yeah, when that commit comes through, you've got the right styling.

jzempel (approved):

PR #202, opened 2020-10-28
feat(content): Improve scannability with punctuation symbols in headings

vsrixyz:
## Description

This was a request by the localization team. I have added the punctuation symbol to each heading so it's more easily scanned. My initial impulse is to put the symbol in parentheses, but I am open to other design solutions!

## Checklist

### DIFF
--- a/src/pages/content/punctuation.mdx
+++ b/src/pages/content/punctuation.mdx
@@ -1,16 +1,18 @@
 ---
 title: Punctuation
+description: >
+  How to use punctuation at Zendesk.
 ---
 
-## Ampersand
+## Ampersand (&)
 
 Avoid ampersands in place of “and.” Don’t use “+” unless you’re opening a coffee shop in Portland.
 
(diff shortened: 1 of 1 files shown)


hzhu (approved):
LGTM

jzempel (commented):
@vsrixyz are you concerned about the impact to in-page permalinks? Also, the parenthesis parentheses feels weird. Not sure I’m feeling this direction.

ginnywood (approved):
Punchy but effective for the goal of scannability.
One alternative I can think of is including the punctuation mark below the header. But that would be more subtle—maybe too subtle. I think this works.

jzempel:
The experts have spoken. 🙌 Thanks y'all.

allisonacs:
Y'all, sorry for opening a can of worms, but this does not meet visual design standards, and I'd like to see this rolled back. We can do better while still improving our users' ability to find what they're looking for.

PR #203, opened 2020-10-29
fix(dropdowns): fix nested menu to highlight correct item

hzhu:
## Description

It looks like the nested menu example has a regression from how it [previously behaved](https://garden-v7.netlify.app/dropdowns/#!/Menu%20usage/7) in the v7 example. The problem is that when a user switches between menus no item is highlighted.

## Detail

The `onStateChange` callback has some incorrect logic which this PR fixes. When navigating between the parent and nested menus, the menu item at index 1 should be highlighted. Also, `'flowers'` is not an item that exists in the example so that has been removed.

## Screenshots

**Before**: The user switches between menus no item is highlighted

--------

**After**: The user switches between menus and the first item (or the previous parent item) is highlighted.

## Checklist


austingreendev (approved):

jzempel (approved):
Nice, thanks @hzhu

PR #206, opened 2020-11-03
feat: add theme provider component documentation

jzempel:
## Description

New pages appear under `Foundations > Theming` nav:

- Palette
- Theme object
- Theme provider
- Utilities

## Detail

Reviewers (@vsrixyz, @rossmoody, @allisonacs) please note that all sections, including API, were documented in this PR rather than auto-generated. As such, it will be good to put :eyes: on all aspects – API sections included. I also made some game-time decisions apart from the previously drafted documentation. Mostly having to do with the segments we were feeling iffy about.

## Checklist


src/examples/utilities/components.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+  },
+  {
+    name: 'withTheme',
+    props: {
+      Component: {
+        required: true,
+        type: 'Object',
+        description:
+          'Takes a component to be wrapped by a higher-order component with an added `theme` prop'
```
vsrixyz, line 166: This one is kind of hard to parse. What does it mean to "take a component"?

src/pages/components/palette.mdx
```diff
@@ -0,0 +1,59 @@
...
+      <li>
+        <Markdown>
+          To style custom components with Garden, access the palette through the
+          [`theme.palette`](/components/theme-object#palette) API or
+          [`withTheme`](/components/utilities#withTheme) utility instead
+        </Markdown>
+      </li>
+      <li>
+        <Markdown>
+          To extend Garden with reusable components that respond to theming, use the
```
vsrixyz, line 30: I think you can lose "together":

To extend Garden with reusable components that respond to theming, use the [`getColor`](/components/utilities#getColor) utility ~together~ with the [`theme.colors`](/components/theme-object#colors) API instead

src/pages/components/theme-object.mdx
```diff
@@ -0,0 +1,236 @@
...
+values are weights used by Garden components.
+
+### iconSizes
+
+The icon sizes object corresponds with the small, medium, and large icons
+provided by Garden's [SVG Icon](/design/icons/installation) library.
+
+### lineHeights
+
+Line heights are used in conjunction with [`fontSizes`](#fontSizes) to define
```
vsrixyz, line 188: You can make this simpler:

Line heights are used ~n conjunction~ with

src/pages/components/utilities.mdx
```diff
@@ -0,0 +1,121 @@
...
+  - theming/src/utils/arrowStyles.ts
+  - theming/src/utils/getColor.ts
+  - theming/src/utils/getLineHeight.ts
+  - theming/src/utils/mediaQuery.ts
+  - theming/src/utils/menuStyles.ts
+  - theming/src/utils/retrieveComponentStyles.ts
+  - theming/src/utils/withTheme.ts
+---
+
+## How to use it
```
vsrixyz, line 17: Should we lose this heading since there is nothing underneath it?

src/pages/components/utilities.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+### Menu and arrow styles
+
+Use menu and arrow styling utilities together to ease the production of
```
vsrixyz, line 21: What does "ease the production" mean? Do we mean: Use menu and arrow styling utilities to make it easier to build Garden-styled menu components?

vsrixyz (commented):
Only a few suggestions

src/examples/utilities/components.ts
jzempel: @vsrixyz I agree. Yet this is parlance that is familiar in React land. Check out https://reactjs.org/docs/higher-order-components.html

src/pages/components/utilities.mdx
jzempel: No. This is standard throughout the component pages and we must retain a valid H1 -> H2 -> H3 for accessibility.

src/pages/components/utilities.mdx
jzempel: Yes. In fact, we can probably simplify to:

> Use menu and arrow styling utilities to build Garden-styled menu components.

src/examples/utilities/components.ts
jzempel: ...although, I think I can do better.

> Gets wrapped by a higher-order component with an added `theme` prop

src/pages/components/utilities.mdx
jzempel: Although, should this be "How to use them" as we are referring to Utilities rather than a single Component?

src/pages/components/theme-object.mdx
```diff
@@ -0,0 +1,235 @@
...
+
+### palette
+
+This object lets you reference, modify, and extend Garden’s entire
+[Palette](/components/palette) (with the exception of `palette.product`
+values, which cannot be customized).
+
+### rtl
+
+The RTL theme object value determines whether components will layout
```
rossmoody, line 200: `lay out`

rossmoody:
Looking so good. Broader topic to consider: Since there is no visual difference between a `code` snippet and a `code` snippet that is also a link, I wonder if it would be better to go with an anchor styling so links don't go unnoticed.

jzempel:
> Since there is no visual difference between a `code` snippet and a `code` snippet that is also a link, I wonder if it would be better to go with an anchor styling so links don't go unnoticed. 

Yesss. I was hoping you would help me make the call on this. Let's work together to determine a direction for this PR and future `website` content.

src/pages/components/theme-provider.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+<CodeExample code={NestingCode}>
+  <Nesting />
+</CodeExample>
+
+### Targeting
+
+The theme's [components](/components/theme-object#components) object can
+override styling for a targeted component type. In this example, Garden
+[Tabs](/component/tabs) are restyled to appear below associated panel
```
hzhu, line 83: It looks like this URL is broken because the path should be "/components/tabs".

src/pages/components/theme-provider.mdx
```diff
@@ -0,0 +1,113 @@
...
+### Default
+
+The typical usage of a theme provided to styled components.
+
+import Default from '../../examples/theme-provider/ThemeProviderDefault.tsx';
+import DefaultCode from '!!raw-loader!../../examples/theme-provider/ThemeProviderDefault.tsx';
+
+<CodeExample code={DefaultCode}>
+  <Default />
+</CodeExample>
```
hzhu, line 36: The examples look good on the website but seem like it's rendered differently on CodeSandbox. Anyone else experiencing this?

Screen Shot 2020-11-03 at 12 43 51 PM

src/pages/components/utilities.mdx
```diff
@@ -0,0 +1,120 @@
...
+are applied must be wrapped in a positioned box (`relative`, `absolute`, or
+`fixed`) that has a `z-index` greater than or equal to zero.
+
+<PropSheet components={COMPONENTS} componentName="arrowStyles" headerName="Parameter" />
+
+### getColor
+
+Get a color for the given hue, shade, and theme. The color is resolved with a
+fallback algorithm that succeeds whether or not the particular shade exists
+within the theme.
```
allisonacs, line 52: This wording is confusing to me as a non-developer. I think what it's trying to say is that `getColor` will provide a color even if the particular shade is not part of the theme. 

Consider being more explicit that the shade doesn't have to be in the theme. I think that's what was confusing to me. Also the word "succeeds." Maybe something like "…the algorithm returns a valid color even if the particular shade doesn't exist in the current theme."

allisonacs (changes requested):
minor content edits

src/pages/components/theme-provider.mdx
jzempel: Ah, I think it's because I'm using `css` rather than `style`. I'll get this fixed.

src/components/MarkdownProvider/components/Configuration.tsx
```diff
@@ -76,7 +77,8 @@ export const Configuration: React.FC<{ reactPackage: IPackage; components: IComp
       <StyledListItem>
         <StyledListItemLabel isBold>Deps</StyledListItemLabel>
         <StyledMono>
-          npm install react react-dom prop-types styled-components @zendeskgarden/react-theming
+          npm install react react-dom prop-types styled-components
+          {!isTheming && ' @zendeskgarden/react-theming'}
```
austingreendev, line 81: Rather than a separate prop, could we use `reactPackage.packageName === '@zendeskgarden/react-theming'` to dynamically apply this value?

src/pages/components/palette.mdx
```diff
@@ -0,0 +1,59 @@
...
+      </li>
+    </ul>
+  </Use>
+  <Misuse>
+    <ul>
+      <li>
+        <Markdown>
+          To style custom components with Garden, access the palette through the
+          [`theme.palette`](/components/theme-object#palette) API or
+          [`withTheme`](/components/utilities#withTheme) utility instead
```
austingreendev, line 25: Should we reference the `ThemeContext` object as that's the "new" approach provided by styled-components?

jzempel: Yeah, probably. The link to `withTheme` gets to that detail. But I was almost wondering if we wanted to ditch documenting `withTheme` on the website?

austingreendev: To support class components I don't think the HOC is going anywhere on the styled-components side. So I guess that's a good reason to leave it around at least.

jzempel: I just elected to remove the `withTheme` clause here. In this context it is relevant to reference the theme object's key, less relevant to determine how to access (styled vs. non-styled Component) the key.

hzhu (approved):

rossmoody (approved):

allisonacs (approved):

PR #205, opened 2020-10-30
feat: new FileUpload documentation page

austingreendev:
## Description

This PR adds the documentation for the `FileUpload` component. The DO/DONT layer isn't merged into abstract yet, but that will be added once available.

## Checklist

- [x] :ok_hand: website updates are Garden Designer approved (add the designer as a reviewer)
- [x] :black_nib: copy updates are approved (add the content strategist as a reviewer)
- [x] :link: considered opportunities for adding cross-reference URLs (grep for keywords)
- [x] :wheelchair: analyzed via [axe](https://www.deque.com/axe/) and evaluated using VoiceOver
- [x] :memo: tested in Chrome, Firefox, Safari, Edge, and IE11

PR #207, opened 2020-11-03
chore: best practice styles

rossmoody:
## Description

This PR temporarily addresses the poor formatting the site is experiencing in relation to the static images in the best practices section. I think the styling of that component in general needs a larger design investigation, but I think this fix would allow the images to format more pleasantly in full width and half width scenarios. 

This is compared against master but the changes reflected on #205  would go [from this](https://share.getcloudapp.com/z8uYRod0) to [this.](https://share.getcloudapp.com/YEuybww4) 

The pages with static content to view this change are:

- Avatar
- Autocomplete
- Checkbox
- Icon button
- Toggle icon button

## Checklist

### DIFF
--- a/src/components/MarkdownProvider/components/BestPractice.tsx
+++ b/src/components/MarkdownProvider/components/BestPractice.tsx
@@ -117,7 +117,6 @@ interface ISectionProps extends ICaptionProps {
   imageAlt?: string;
   imageHeight?: number;
   imageWidth?: number;
-  imageIsSquare?: boolean;
 }
 
 export const Section: React.FC<ISectionProps> = props => {
@@ -125,8 +124,7 @@ export const Section: React.FC<ISectionProps> = props => {
     const imageStyles = {
       width: props.imageWidth,
       height: props.imageHeight,
-      maxWidth: props.imageIsSquare ? 160 : undefined,
-      maxHeight: 160
+      maxWidth: 320
     };
 
     return (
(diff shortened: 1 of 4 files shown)


rossmoody:
@steue  @austingreendev If we think this update is a valid temporary fix, I can add the FileUpload image we omitted in the other PR as part of this.

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -117,16 +117,14 @@ interface ISectionProps extends ICaptionProps {
...
   imageWidth?: number;
-  imageIsSquare?: boolean;
 }
 
 export const Section: React.FC<ISectionProps> = props => {
   if (props.imageSource) {
     const imageStyles = {
       width: props.imageWidth,
       height: props.imageHeight,
-      maxWidth: props.imageIsSquare ? 160 : undefined,
```
jzempel, line 128: I think you'll still want the `imageIsSquare`, but set to `320`, in order to prevent image skew from a `height` set that conflicts with a `maxWidth`. Or am I missing something?

jzempel (commented):
The following page images are looking overly large:
- Autocomplete
- Icon button
- Toggle icon button

I think we need to hear more about the design intent before we put in guardrails to protect against super-sized images.

src/components/MarkdownProvider/components/BestPractice.tsx
rossmoody: Oh, good spot. This update may be ill advised. I didn't see any images with width or heights explicitly set so the thinking here was that the image would proportionally fill the div up to 320px wide. Perhaps a more appropriate solve would be to set the height and width explicitly but I think images with details won't work below 160px. 

Either way I think we're seeing that this does indeed need more investigation so I'll can this one in favor of design investigation.

PR #213, opened 2020-11-18
feat(typography): add Blockquote component example, configuration, and API

jzempel:
## Description

See the `/components/typography#blockquote` URL for review.

## Checklist


steue (approved):
looks good to me! the garden related quotes fit well. i wonder if v can sprinkle some magic spice?

src/pages/components/typography.mdx
```diff
@@ -16,6 +17,17 @@ components:
 
 ## How to use it
 
+### Block quotation
```
allisonacs, line 20: Is this the right terminology? I've typically heard these referred to simply as `block quote`.

src/pages/components/typography.mdx
```diff
@@ -16,6 +17,17 @@ components:
...
+
+Use Blockquote to represent a body of text that is from another source.
```
allisonacs, line 22: Another use case is to bring special emphasis to content found within the body. That's editorial, and we may never see anything like that in product… should it be called out anyways, or would it just cause too much confusion? @steue @vsrixyz

src/pages/components/typography.mdx
jzempel: @allisonacs please see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote where we learn this is the "Block Quotation" element. But I'm fine to update, if necessary.

allisonacs: huh, TIL. 

I understand wanting to mirror the standard, but it feels like esoteric language to me. My personal vote is to use a more colloquial term for it, but I get that I am kind of picking/choosing where I want to apply that standard. Up to you 🙂

jzempel: colloquial is king 👑

hzhu (approved):

src/pages/components/typography.mdx
jzempel: Merging per offline convo with @steue. Happy to adjust copy under a separate PR.

PR #226, opened 2020-12-03
feat: add full-text search

jzempel:
## Description

Adds Algolia Docsearch to the website 🎉 .

## Detail

I tracked styling overrides closely with Garden's `Menu`. But there's plenty of room for more customization, if called for by design. 

- Use https://docsearch.algolia.com/demo/ to view DocSearch without any customizations.
- Note the projects listed on this page – https://docsearch.algolia.com/ – to view and compare other customizations.
- re: **copy**, what do we think of the "Search..." and "kale-600 hex value" placeholder text?

@vsrixyz & @allisonacs can you review the search field placeholder text? If we decide to go with an example rather than a directive on the home screen hero, we probably need to select text that results in a valid search. The original design called for "What is kale-600?" which returns 0 valid results.

### Reference

Current DocSearch Garden config: https://github.com/algolia/docsearch-configs/blob/master/configs/garden_zendesk.json

## Checklist

### DIFF
--- a/src/layouts/Home/components/Search.tsx
+++ b/src/layouts/Home/components/Search.tsx
@@ -117,7 +117,7 @@ export const Search: React.FC = () => {
                     }
                   `}
                 >
-                  <SearchInput id="algolia-docsearch-home" placeholder="kale-600 hex value" />
+                  <SearchInput id="algolia-docsearch-home" placeholder="how to use the stepper" />
                 </div>
               </div>
             </Col>
--- a/src/layouts/Home/components/Search.tsx
+++ b/src/layouts/Home/components/Search.tsx
@@ -11,10 +11,9 @@ import Img from 'gatsby-image';
 import { css, ThemeProps, DefaultTheme } from 'styled-components';
 import { getLineHeight, mediaQuery } from '@zendeskgarden/react-theming';
 import { Grid, Row, Col } from '@zendeskgarden/react-grid';
-// import { MediaInput } from '@zendeskgarden/react-forms';
 import { LG } from '@zendeskgarden/react-typography';
-// import { ReactComponent as SearchStroke } from '@zendeskgarden/svg-icons/src/16/search-stroke.svg';
 import MaxWidthLayout from 'layouts/MaxWidth';
+import { SearchInput } from 'layouts/Root/components/SearchInput';
 
 const headerStyling = (p: ThemeProps<DefaultTheme>) => {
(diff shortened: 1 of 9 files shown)


jzempel:
:sob: the Algolia Docsearch wrapper isn't getting applied to the production build. Dropping into "Draft" while I look for a solution.

src/layouts/Root/components/SearchInput.tsx
```diff
@@ -0,0 +1,186 @@
...
+    docsearch: any;
+  }
+}
+
+export const SearchInput = React.forwardRef<HTMLInputElement, HTMLAttributes<HTMLInputElement>>(
+  ({ id, ...props }, ref) => {
+    useEffect(() => {
+      if (typeof window !== 'undefined' /* not SSR */ && window.docsearch) {
+        window.docsearch({
+          apiKey: 'c1ab842b2bfd6f3f7104a8e903c615bf',
```
rossmoody, line 158: Any reason to env variable this in Netlify?

jzempel: Nah, this is how the cool kids [do it](https://github.com/reactjs/reactjs.org/blob/master/src/components/LayoutHeader/DocSearch.js). It's a read-only key, so no worries about leaked index security.

hzhu (approved):

austingreendev (approved):

src/layouts/Home/components/Search.tsx
```diff
@@ -118,12 +117,8 @@ export const Search: React.FC = () => {
                     }
                   `}
                 >
-                  <MediaInput
-                    placeholder="What is kale-600?"
-                    aria-label="Garden search"
-                    start={<SearchStroke />}
-                  />
-                </div> */}
+                  <SearchInput id="algolia-docsearch-home" placeholder="kale-600 hex value" />
```
allisonacs, line 120: since we don't want to encourage people to use the hex vales maybe we change this to: "how to use the stepper"

As an aside, it could be interesting to have this rotate between a bunch of possible placeholders; could also be a subtle way to advertise new components or patterns.

allisonacs (changes requested):
Minor copy update

src/layouts/Home/components/Search.tsx
jzempel: I was thinking the same thing 👍 Thanks, @allisonacs

allisonacs (approved):

PR #232, opened 2021-01-09
chore: add component @extends clauses

jzempel:
## Description

- bump to the latest version of `react-components`
- fill out `<Component>` and `<PropSheet>` data for all components and subcomponents
- tweak a few examples with fixes/improvements

Leaving in DRAFT until we settle on a designed solution. @allisonacs the `@extends` clause we see throughout is a component that can be designed however we see fit with regard to spacing, styling, etc.

Screen Shot 2021-01-08 at 4 51 15 PM

## Detail

✅ ~zendeskgarden/react-components#983 will provide data for any missing @extends clauses (Autocomplete, Dropdown, etc).~

## Checklist

- [ ] :link: considered ~opportunities for adding cross-reference URLs (grep for keywords)~


src/examples/forms/file-upload/Disabled.tsx
```diff
@@ -33,9 +33,9 @@ const Example = () => {
             {isDragActive ? (
               <span>Drop files here</span>
             ) : (
-              <span>Choose a file of drag and drop here</span>
+              <span>Choose a file or drag and drop here</span>
             )}
-            <Input {...getInputProps()} disabled />
+            <input {...getInputProps()} disabled />
```
austingreendev, line 38: With this using a native `input` are the correct aria attributes being applied for the label to be read by assistive devices?

austingreendev (changes requested):
Overall I don't feel good about us maintaining separate `component.ts` files within the website repo. Is there any way for us to get this information from standard JSDoc comments?

I see these falling out of date very quickly since they're not tied to the actual source.

src/examples/forms/file-upload/Disabled.tsx
jzempel: I intended to call this out post-draft – I'm glad you caught it. Yes, this does break the association and needs to be reverted (and subsequently added to the `FileUpload` page imports).

jzempel:
@austingreendev my in-depth research shows there is no way to obtain subcomponent documentation via existing docgen libraries. I think the team had this discussion via Slack. While there is, admittedly, some brittleness, I believe this is our best option going forward. It will be part of the Maintainers responsibility to keep these docs in-sync as new versions of `react-components` are merged (and in the future, auto-merged with manual approval).

jzempel:
> falling out of date very quickly

Over the course of Garden, how often have subcomponent APIs changed? I think the answer is almost never.

austingreendev (approved):

allisonacs:
For formatting, I'm feeling like the code blocks are overly heavy, since these strings are very long and all one word. So I propose moving closer to [FluentUI](https://developer.microsoft.com/en-us/fluentui/#/controls/web/button#implementation), where the word "Extends" is in our normal body copy, and the tag itself is just in monotype at 13px, grey-700.

allisonacs (approved):

PR #237, opened 2021-01-28
feat: add new buttons pattern

austingreendev:
## Description

This PR re-introduces the Patterns IA with the `Buttons` pattern. It includes a live loading example along with best practices and a static image example.

## Checklist


src/pages/patterns/index.mdx
```diff
@@ -0,0 +1,17 @@
+---
+title: Design overview
+description: >
+  Foundations for creating purposeful UI.
```
jzempel, line 4: @allisonacs, @ctancinco I think this will need some work.

src/pages/patterns/buttons.mdx
```diff
@@ -0,0 +1,48 @@
+---
+title: Buttons
+description: >
+  TODO BEED TO FIGURE OUT DESCRIPTION
```
austingreendev: @allisonacs, @ctancinco This section also needs to copy help.

ctancinco: Looks fine to me. 

Haha, I'll review these today.

ctancinco: You have an existing description on the site now: “Buttons let users take action quickly. Use a Button to enable actions or decisions that are important to a user's workflow.” If you don't want to reuse that, we could just edit this down to: "Buttons highlight and enable an important action or decision within a workflow." LMK if you wanted to message something differently.

src/pages/patterns/index.mdx
ctancinco: Let me know what's working/not working and I can give a few options.

jzempel: @ctancinco this was copied from the existing [Design overview](https://garden.zendesk.com/design) page. This new "Patterns overview" will need its own tagline description that aligns with the other top-level pages.

ctancinco: Got it, thank you. Pattern overview options:
- Commonly used components and how they’re used within the product
- Commonly used components for better UX
- Product problems and the components that love them
- Existing design solutions for a more consistent UX

jzempel: - "Commonly" feels ... common.
- "problems" feels ... problematic.
- "solutions" feels right.

Solutions feels more aligned with content:principles, design:foundations, component:instructions.

What do you think, @allisonacs? Or should we riff offline?

src/pages/patterns/buttons.mdx
austingreendev: @jzempel @allisonacs what is your decision on this?

src/pages/patterns/index.mdx
allisonacs: +1, I like solutions. Let's go with that.

src/pages/patterns/buttons.mdx
allisonacs: @ctancinco do we need to differentiate the Component overview from the pattern overview?

Patterns are how you put these components together, how you use them properly… that certainly overlaps with what we have/your suggestion, but doesn't imply that the article tells you how to do that / tells you best practices. 

I'm also mindful that these will all be housed in the "patterns" section itself, so battering people with phrasing like "and here are the best practices for that" will be repetitive and maybe not necessary. 

What do you think?

src/pages/patterns/index.mdx
jzempel: Probably helpful to note that every other hi-level IA page description is reflected on the home page. I would expect this to be the same.

austingreendev: I've updated this to `Existing design solutions for a more consistent UX`. Thanks for the help.

src/pages/patterns/buttons.mdx
ctancinco: @allisonacs Sorry, I didn't see we were talking pattern vs component. Yes, it should be treated differently than components but it has to be shorter than the pattern overview. 

Option 1: Thesis statement. Distill the philosophy behind button design into one line. Examples: 

- There are different buttons for different needs
- Understand the intention behind the action
- Know why they're pushing your buttons

Option 2: best practices/how-to but, as you mentioned, will get repetitive once you add more patterns:

- Tips for better button design
- How to design better buttons

LMK if any of these are closer to the mark.

allisonacs: > Know why they're pushing your buttons

I love that. 

When we look at the component description overviews, they've got a period at the end, hopefully we can do that here too.

ctancinco: absolutely: "Know why they're pushing your buttons."

austingreendev: Updated. Thanks for the help.

src/pages/patterns/buttons.mdx
```diff
@@ -0,0 +1,113 @@
...
+## Size
+
+The default button size is medium. In screens that have more information density, use small
+buttons. Large buttons are mostly used in onboarding. Buttons and text inputs within the same
+flow should be the same size.
+
+<BestPractice>
+  <Do imageIsSquare imageSource={props.data.buttonsSizeDo.childFile.childImageSharp.fluid} />
+  <Dont imageIsSquare imageSource={props.data.buttonsSizeDont.childFile.childImageSharp.fluid} />
+</BestPractice>
```
austingreendev, line 31: @jzempel I'm trying to figure out the best logic for these images, but they aren't rendering to a valid height. Have we ran into tall images yet?

jzempel: Shouldn't the `imageIsSquare` attribute be removed? Beyond that, I would guess the underlying component hasn't been stressed with this type of layout and could need to be updated.

austingreendev: Correct, that shouldn't be needed. Was just trying to resolve some of the resizing issues. I'll update the `BestPractices` styling logic.

austingreendev: I've updated the hard-coded `maxHeight` to only be applied with square images.

src/pages/patterns/index.mdx
```diff
@@ -0,0 +1,17 @@
+---
+title: Patterns overview
+description: >
+  Existing design solutions for a more consistent UX
```
jzempel, line 4: @ctancinco, @allisonacs for all other pillars (Content, Design, Components) this description matches the home page 1:1, including no-period. Should the same thing happen here or on the home page?

src/pages/patterns/buttons.mdx
```diff
@@ -0,0 +1,113 @@
...
+page or workflow. Buttons use visual signifiers such as shape, color, and size to show that
+a user can interact with them.
+
+## Interface
+
+**Primary buttons** are for the most important action in a given view. There is typically
+one call-to-action on a page or view. **Default** buttons are for secondary actions.
+**Basic buttons** are for repetitive or deemphasized actions in a given view.
+
+![Tone matrix map](abstract:patterns-buttons-type-usage.png)
```
jzempel, line 20: This image is grossly oversized.

src/pages/patterns/buttons.mdx
```diff
@@ -0,0 +1,113 @@
...
+Buttons are common and vital elements because they help make content interactive across the
+internet. They allow a user to take actions with a single click and are placed throughout a
+page or workflow. Buttons use visual signifiers such as shape, color, and size to show that
+a user can interact with them.
+
+## Interface
+
+**Primary buttons** are for the most important action in a given view. There is typically
+one call-to-action on a page or view. **Default** buttons are for secondary actions.
+**Basic buttons** are for repetitive or deemphasized actions in a given view.
```
jzempel, line 18: Why would these treatments be bold rather than links to the relevant sections?

src/pages/patterns/buttons.mdx
```diff
@@ -0,0 +1,113 @@
...
+## Size
+
+The default button size is medium. In screens that have more information density, use small
+buttons. Large buttons are mostly used in onboarding. Buttons and text inputs within the same
+flow should be the same size.
+
+<BestPractice>
+  <Do imageSource={props.data.buttonsSizeDo.childFile.childImageSharp.fluid} />
+  <Dont imageSource={props.data.buttonsSizeDont.childFile.childImageSharp.fluid} />
+</BestPractice>
```
jzempel, line 31: This component was never designed to be shown without inline description text. Also, it seems important to give the user a helpful hint similar to https://garden.zendesk.com/components/avatar#select-the-right-shape-for-the-job

src/pages/patterns/buttons.mdx
```diff
@@ -0,0 +1,113 @@
...
+  <Do imageSource={props.data.buttonsContentVerbDo.childFile.childImageSharp.fluid}>
+    <ul>
+      <li>Use labels specific to the action they perform</li>
+      <li>Communicate a single action per button</li>
+    </ul>
+  </Do>
+  <Dont imageSource={props.data.buttonsContentVerbDont.childFile.childImageSharp.fluid}>
+    <ul>
+      <li>Use general button labels that might cause confusion</li>
+      <li>Use labels that need more context</li>
```
jzempel, line 53: The mismatch of putting "Use" imperative text under a "Don't" category is confusing. What am I actually supposed to do?

jzempel (changes requested):
@m-lai, @ctancinco, @allisonacs a few notes...

- the overall structure of the page doesn't feel as thought-out, compelling, or winsome as current Component pages (i.e. "How to use it", "How to use it well") and I feel like there's not a clear map or user expectation for future patterns
- I don't think this page should publish until thought has been given to existing content that should be linked to and content in here that should be referenced from existing content

p.s. the actual loading button example is super useful :wink:

src/pages/patterns/buttons.mdx
```diff
@@ -0,0 +1,113 @@
+---
+title: Buttons
+description: >
+  Know why they're pushing your buttons.
```
jzempel, line 4: Please apply a curly quote here as the gatsby plugin doesn't hit this portion of MDX.

jzempel (changes requested):

m-lai:
@jzempel Agree with your sentiments. Some early thoughts I have are:

• We currently have a template of Overview, Interface, Content, Interactions. Regardless of whether these get tweaked/changed/improved, I do think it would be useful to have these sections display on either the sidenav under the parent Pattern name, or on the main page itself near the top. This would give users a better overview of the content or specific things they're looking for.

• Providing standard header images at the top of each pattern could help anchor users and give more context on what to expect. Sometimes the first image you see in documentation is not early on, or specifically a general image related to the pattern.

• Are we able to allow static images to be clicked on to expand view or at least open in a new window? Some of these modal images are v hard to see

• The single modal image is oversized. We should try to cap images at ~600px or so to try to make it feel more intentional

src/pages/patterns/buttons.mdx
austingreendev: Rather than have us manually enter these, I've added a customization to our page creation logic that runs all frontmatter title and description fields through https://smartquotes.js.org/

This also corrected some rogue quotes in our component pages.

Buttons component before:
before

Buttons component after:
after

src/pages/patterns/buttons.mdx
jzempel: @austingreendev could the Abstract plugin be structured to receive `@XxY` syntax similar to something used by https://placeholder.com/? So `abstract:image.png@300x150` would be 300px width, 150px height. And `abstract:image.png@300` would be 300px width, auto height.

src/pages/patterns/buttons.mdx
jzempel: Awesome :tada:

src/pages/patterns/buttons.mdx
austingreendev: I'm not sure if that information would be possible to provide from there. We're really only updating the `src` URL with that remark plugin at this time. All the sizing and other responsive customizations happen in `gatsby-remark-images`.

If we want to get specific on a per-image basis we may want to reference the Abstract asset directly in a page query and apply customizations to an `Img` element.

src/pages/patterns/index.mdx
allisonacs: Yes. Let's update the homepage to match this copy, it feels more on brand.

src/pages/patterns/buttons.mdx
jzempel, line 20: The alt text needs attention here.

austingreendev:
Closing this for now. Thanks, everyone!

jzempel:
Lot's of good work in here, @austingreendev. Looking forward to recovery in the near future.

PR #255, opened 2021-03-25
feat(forms, dropdowns): add hidden label examples to component pages

jzempel:
## Description

Improve new `hidden` prop discovery with supporting examples to the following component pages:
- Autocomplete
- Multiselect
- Select
- Input
- Range
- Textarea

All examples are essentially copies of the existing default, with a modification to the `Label` component.

## Checklist


jzempel:
@austingreendev when this merges, you would have the v8.35.0 features needed for the notification example.

hzhu (approved):

austingreendev (approved):

allisonacs (approved):
Approved, with one change request. Please also add this to the multi-thumb range. While we hopefully wouldn't see that used, it feels odd to be that we would demonstrate this behavior on the single thumb version, but not the multi.

PR #260, opened 2021-04-06
feat(patterns): new buttons documentation

jzempel:
## Description

_In draft as Patterns discussions continue_

## Checklist


src/pages/patterns/examples/button/Interaction.tsx
```diff
@@ -0,0 +1,54 @@
...
+const InteractionExample = () => {
+  const themeContext = useContext(ThemeContext);
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    const buttonElement = buttonRef.current;
+
+    if (!isLoading && buttonElement) {
+      buttonElement.style.width = `${buttonElement.clientWidth}px`;
```
mtomcal, line 23: Why not `useState` to manage a `style` prop for `width`?

hzhu: That's a good question!

I think it's because `width` isn't state — it doesn't drive any React UI. Rather, it's a DOM CSS property that's being imperatively set in a side effect.

The `useRef` hook can be used for accessing a DOM node or for persisting values across renders. One way I think about `useRef` vs `useState` is that they both persist value across renders — the only difference is that `useState` returns a state updater function to update state, which drives React UI. In this case, there is no React UI (JSX portion) that depends on the `width`.

Having said that, `useState` could be used, and that can be a valid alternative. But I think what @jzempel has is idiomatic for what is being done.

mtomcal: Philosophical: In many ways mutating a ref could be breaking the contract with React prop driven model, reverting back to the days of jQuery where mutation of one element and other components not being informed. With everything driven via props from a state update, we maintain that model rather than mutating in a side effect. Using `ref`s for accomplishing things outside of React's control/API (aka focus management) or using refs for read-only purposes makes complete sense to me but mutating styling when we have a prop driven API seems very anti-pattern in the React world and would not scale. Ultimately, React seems to be one big state machine managing the DOM and that includes various states of UI including width changes. I would hope our documentation advertises the React way to accomplish things.

mtomcal: > Avoid using refs for anything that can be done declaratively.

https://reactjs.org/docs/refs-and-the-dom.html#when-to-use-refs

We can do things declaratively via `style` prop.

hzhu: I disagree. I also disagree that what @jzempel has reverting back to jQuery model. Why would using `useState` be such a big deal in this case? A side effect (use effect) must exist anyways to read the `width`. So you're suggest adding a `useState` and triggering another render to have React re-render the UI? Not necessarily wrong, but I think that's unnecessary.

jzempel: Probably not worth discussing too hard. This draft PR is not eligible for merge. The core will be used as the baseline for the website "Patterns" pillar. But the buttons documentation is not design-sanctioned for publish.

jzempel:
displaced by #341

PR #270, opened 2021-05-12
chore: update Garden components to v8.37.0

jzempel:
## Description

- Add `isNeutral` treatment to `InputGroup` buttons
- Order `InputGroup` examples alphabetically
- Wrap the `Radio` validation example with a `Fieldset`
- Fix `Tag` close example

## Checklist


hzhu (approved):

src/examples/forms/input-group/Disabled.tsx
```diff
@@ -17,7 +17,9 @@ const Example = () => (
         <Label>Plant name generator</Label>
         <InputGroup>
           <Input value="Sheepberry" disabled />
-          <Button disabled>Germinate</Button>
+          <Button disabled isNeutral>
+            Germinate
```
allisonacs, line 21: With the visual change, the word “Germinate” no longer reads as the action. Let’s change the copy to “Generate”.

src/examples/forms/input-group/Disabled.tsx
```diff
@@ -17,7 +17,9 @@ const Example = () => (
         <Label>Plant name generator</Label>
         <InputGroup>
           <Input value="Sheepberry" disabled />
```
allisonacs, line 19: Tied to the comment about the button copy, maybe we shouldn’t show this field as filled out in the demo.

src/examples/forms/input-group/Size.tsx
```diff
@@ -34,7 +36,7 @@ const Example = () => (
         <Label>Plant name generator</Label>
         <InputGroup isCompact>
```
allisonacs, line 37: I know this is expected, based on how our compact fields are vs. our small buttons, but I can’t help but think this feels wrong. The type sizes between the field and the button are different. :/ 

image

src/examples/forms/input-group/Disabled.tsx
allisonacs: (this should be true in all the examples)

src/examples/forms/radio/Validation.tsx
```diff
@@ -31,8 +30,8 @@ const Example = () => {
   return (
     <Row justifyContent="center">
       <Col size="auto">
-        <MD isBold>Choose a growth type</MD>
-        <div role="group" aria-label="Choose a growth type">
+        <Fieldset>
```
allisonacs, line 33: Should we care too hard that the version using the `Fieldset` shows the radios spaced at the (now) proper 8px margin, but all the others, where no `Fieldset` is present have them stacked with no margin? 

image

allisonacs (changes requested):

src/examples/forms/input-group/Size.tsx
jzempel: Hmm, interesting – now I can't unsee that. This would need to be a component design issue for a separate PR on `react-components` if we decide the change is needed.

src/examples/forms/radio/Validation.tsx
jzempel: Yeah, I do care. But as I attempted to align all the examples I came to the belief that, if we decide to expose a treatment, then all the toggle-style examples need to be cut down to singles and we should have a separate new `<Fieldset>` page showing how various groupings work out. In other words, the situation needs design consideration.

jzempel: (also note that the change you're suggesting turns all of the existing semi-bold labels into regular weight)

src/examples/forms/input-group/Disabled.tsx
jzempel: updated.

src/examples/forms/input-group/Disabled.tsx
jzempel: updated.

src/examples/forms/input-group/Size.tsx
allisonacs: @nyckie @jon-kerwin what do y’all think here?

src/examples/forms/radio/Validation.tsx
allisonacs: I think that’s fair (a new page with cross linking).

allisonacs (approved):

Issue #282, opened 2021-06-15
Images are not visible on Internet Explorer 11


hzhu:
## Expectations

Images should be visible on Internet Explorer 11.

## Reality

Images are not visible on Internet Explorer 11.

<img width="2557" alt="Screen Shot 2021-06-15 at 12 35 23 PM" src="https://user-images.githubusercontent.com/1811365/122113245-af240180-cdd6-11eb-90a5-332a0b41fe28.png">

## Steps to Reproduce

1. Open Internet Explorer 11
1. Navigate to https://garden.zendesk.com/
1. Notice the images do not show up

### Fine Print

- URL: https://garden.zendesk.com/
- Version: N/A
- Browsers: IE11

jzempel:
From `gatsby-image`...

> `gatsby-image` is now backed by the newer `<picture>` tag. This newer standard allows for media types to be chosen by the browser without using JavaScript. It also is backward compatible to older browsers (IE 11, etc).

In fact, IE11 [does not support](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#browser_compatibility) the `<picture>` tag. But I don't think we need to be concerned about this for the website. I'd recommend closing this as no-plans-to-fix.

PR #283, opened 2021-06-15
feat: create gatsby plugin to source images from figma

hzhu:
## Description

Create a Gatsby plugin to source Figma image assets.

## Detail

The website is sourcing image assets from Abstract. However, assets will fully migrate over to Figma sometime in Q4. This PR sets up the foundation of the Figma API integration. To keep the PR focused on the sourcing, it only migrates three images on the home page to be sourced from Figma — a separate PR will be dedicated to migrating the rest of the image assets.

## Screenshot

The three images indicated by the red arrows are now sourced from Figma instead of Abstract.

Screen Shot 2021-06-15 at 12 53 21 PM

## Notes

There is an _existing_ issue with images in IE11 as outlined in https://github.com/zendeskgarden/website/issues/282.

## Checklist


hzhu:
🛑 ⚠️ 🚧 

Hold review. Code should be good, but Sketch -> Figma migration is corrupt, resulting in thousands of unnecessary images. This causes a memory issue in the CI builds.

plugins/gatsby-source-figma/gatsby-node.js
```diff
@@ -0,0 +1,115 @@
...
+  const baseUrl = 'https://api.figma.com/v1';
+  const headers = {
+    'Content-Type': 'application/json',
+    'X-FIGMA-TOKEN': configOptions.figmaApiToken
+  };
+
+  try {
+    const response = await fetch(`${baseUrl}/files/${configOptions.fileId}`, { headers });
+    const { document } = await response.json();
+    const nodes = createNodeList(document).filter(node => node.type === 'FRAME');
```
hzhu, line 47: The alternative to generating a list of nodes would be to directly traverse the tree (`document`). The drawback of directly traversing the tree is that image URLs would need to be fetched one at a time, resulting in `N` network requests.

Instead, I went with generating a list of nodes and using the Figma API to fetch all the image URLs in a single request.

hzhu, line 47: The alternative to generating a list of nodes would be to directly traverse the tree (`document`). The drawback of directly traversing the tree is that image URLs would need to be fetched one at a time, resulting in `N` network requests where `N` is the number of Figma frames. 

Instead, I went with generating a list of nodes, which allows for fetching all the image URLs in a single request.

plugins/gatsby-source-figma/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "private": true,
+  "name": "gatsby-source-figma",
```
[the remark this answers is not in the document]
hzhu: Hmm, I'll take a look at that library! The naming was mostly based off of the current Abstract plugin we wrote called `gatsby-source-abstract`.

src/layouts/Home/components/Foundation.tsx
```diff
@@ -83,8 +83,8 @@ export const Foundation: React.FC = () => {
   const { contentImage, componentsImage, designImage } = useStaticQuery(
     graphql`
       {
-        contentImage: abstractAsset(layerName: { eq: "home-pillars-content" }) {
-          childFile {
+        contentImage: figmaAsset(name: { eq: "home-pillars-content" }) {
```
jzempel, line 86: There is a right-left jitter on the progressive load of this image that can be seen on a hard page refresh. Let's see if we can get this right.

jzempel (changes requested):

src/layouts/Home/components/Foundation.tsx
bethanybeech: I'm also seeing the jitter. Just to be sure, can someone send me the source Figma file so I can see if there's a problem with the way the artboard is set up there?

hzhu: The assets are being pulled from Figma file `HifWgSrdeTlMMXUrcnrAAH`.

I only notice the jitter on the "home-pillars-content" image, but not the other two images. I replaced the avatar "Do" image locally, and it also doesn't seem to have a jitter. So maybe the jitter only affects the "home-pillars-content"?

@bethanybeech @james-sann, does it look like there's a problem with that particular asset? Maybe we can re-create that asset and I can try that instead?

james-sann: @hzhu Re-created the `home-pillars-content` frame. Let me know how you go 🤞

I placed the older version to the side of it in the Figma file and named it `home-pillars-content-old`

Just a side-note on these images and the website conversion. It appears we are surfacing raster images and not SVGs? (at least, that's what I see in Chrome and Safari) Is that correct? When I zoom into those images on the website, I get pixelations but typically SVG doesn't do that? Or does our plugin only support raster images? 
image

jzempel: If we can't get these to (at a minimum) render as well as what we currently see on production, it might make sense to add and reference SVGs directly from the `/website/static` folder.

hzhu: Thanks @james-sann for re-creating the frame. 

Unfortunately, the image still has a jitter. It seems like the placeholder image is currently generated with the option of "TRACED_SVG", and maybe there is an inaccuracy with the tracing. Using the default option "DOMINANT_COLOR" addresses the jitter, but removes the tracing feature. Would that be acceptable? Here's a [demo of that](https://ecstatic-benz-474ab9.netlify.app/).

Alternatively, I can put a SVG into the `/website/static` folder for `home-pillars-content`.

hzhu: The `home-pillar-*` images are now powered by static SVGs (exported from Figma). 

Since the PR aims to test Figma API integration, I've decided to migrate at least one image that sources from Figma directly, so this PR now sources the `home-hero-logo`. It looks like there are some issues on the Figma side. There's an extra white background, and it seems like a layer might be missing. This can be visible on the latest "View deployment" link.

@james-sann would you be able to check out what might be wrong with `home-hero-logo`?

james-sann: Hi @hzhu. Try it now 🤞 

I removed these settings from Figma and it should remove any background color fill getting exported (you can see the preview within Figma's export panel).

image

src/layouts/Home/components/Foundation.tsx
```diff
@@ -16,32 +15,29 @@ import MaxWidthLayout from 'layouts/MaxWidth';
...
+const imgStyles = {
+  width: 350,
+  minWidth: '100%',
+  maxWidth: '100%',
+  height: 266,
+  minHeight: '100%',
+  maxHeight: '100%',
+  zIndex: -1
+};
+
```
jzempel, line 27: Looks like these styles were extracted from the hero image and now incorrectly applied to the SVGs. Either way, I'm not sure they are needed.

src/layouts/Home/components/Search.tsx
```diff
@@ -35,8 +35,8 @@ export const Search: React.FC = () => {
   const { bannerImage } = useStaticQuery(
     graphql`
       {
-        bannerImage: abstractAsset(layerName: { eq: "home-hero-logo" }) {
-          childFile {
+        bannerImage: figmaAsset(name: { eq: "home-hero-logo" }) {
+          remoteImage {
```
jzempel, line 39: Is this API variation desirable? Seems like it will be an unwanted, less explicit `image` -> `children` impact for the rest of the codebase.

jzempel (changes requested):
The `svg` files need to be cleaned-up. Please pull down the latest `zendeskgarden/svg-icons` and run the `yarn format:optimize` script. Ex:

```sh
svg-icons> yarn format:optimize ../website/src/images
```

From there, the files will need added manual cleanup as it appears that Figma is adding some phantom shapes. Please remove unwanted `rect` and `path` elements, fix up rect size, remove root `fill`, and typically `fill-rule` attributes can be removed. Fore example, the design SVG...

**before**

```svg
<svg width="288" height="220" viewBox="0 0 288 220" fill="none" xmlns="http://www.w3.org/2000/svg">
<rect width="288" height="220" fill="white"/>
<rect width="288" height="220" fill="#EDE0CF"/>
<path fill-rule="evenodd" clip-rule="evenodd" d="M64 30.0149C64.2759 30.013 64.55 30 64.8259 30C152.67 30 223.893 102.007 224 190H64V30.0149Z" fill="#00363D"/>
<path fill-rule="evenodd" clip-rule="evenodd" d="M84 94.0071C84.1311 94.0062 84.2612 94 84.3923 94C126.118 94 159.949 128.203 160 170H84V94.0071Z" fill="#228F67"/>
</svg>
```

**after**

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="288" height="220" focusable="false" viewBox="0 0 288 220">
  <rect width="100%" height="100%" fill="#EDE0CF"/>
  <path fill="#00363D" d="M64 30.0149C64.2759 30.013 64.55 30 64.8259 30C152.67 30 223.893 102.007 224 190H64V30.0149Z"/>
  <path fill="#228F67" d="M84 94.0071C84.1311 94.0062 84.2612 94 84.3923 94C126.118 94 159.949 128.203 160 170H84V94.0071Z"/>
</svg>
```

src/layouts/Home/components/Foundation.tsx
```diff
@@ -163,8 +119,15 @@ export const Foundation: React.FC = () => {
...
-                  image={contentImage}
-                />
+                >
+                  <StaticImage
+                    alt=""
+                    placeholder="tracedSVG"
+                    src="../../../images/home-pillars-content.svg"
+                    tracedSVGOptions={{ color: '#478d6b', background: '#F6F4F4' }}
+                    imgStyle={imgStyles}
+                  />
```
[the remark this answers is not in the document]
hzhu: Using an `img` tag would remove the Gatsby image's [placeholder](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-plugin-image/#placeholder) feature. Rather than using static images, I think using the `gatsby-image-plugin` placeholder option of `DOMINANT_COLOR` would be better for the home page pillar images. It would be a three line change (deletion) and keep the change simplified. Also, the aesthetics would more closely map to what's in production in that there is a placeholder for each home page pillar image.

src/layouts/Home/components/Foundation.tsx
hzhu: @james-sann, that seemed to have worked, thanks!

gatsby-config.js
```diff
@@ -11,6 +11,9 @@ const path = require('path');
 
 require('dotenv').config();
 envalid.cleanEnv(process.env, { ABSTRACT_TOKEN: envalid.str() });
+envalid.cleanEnv(process.env, { FIGMA_TOKEN: envalid.str() });
```
jzempel, line 14: This can simply be added to the object above...

`{ ABSTRACT_TOKEN: envalid.str(), FIGMA_TOKEN: envalid.str() }`

jzempel (changes requested):

plugins/gatsby-transformer-garden-svg/gatsby-node.js
```diff
@@ -39,6 +39,10 @@ exports.onCreateNode = async ({
   createNodeId,
   createContentDigest
 }) => {
+  if (node.internal.type === 'StaticImage') {
+    return;
+  }
+
```
jzempel, line 45: Is this still needed?

jzempel (approved):
I will plan to address the `Content` image jitter under a separate SVG "spritemap" PR.

plugins/gatsby-transformer-garden-svg/gatsby-node.js
hzhu: Ah, this needs to be removed.

jzempel (changes requested):
Oops, this PR has the wrong (not color-corrected and different) image for the hero.

src/layouts/Home/components/Foundation.tsx
hzhu: @james-sann oops, looks like the home hero image successfully removed the white background, but some layers seem to be missing.

gatsby-config.js
```diff
@@ -10,7 +10,9 @@ const fs = require('fs');
 const path = require('path');
 
 require('dotenv').config();
-envalid.cleanEnv(process.env, { ABSTRACT_TOKEN: envalid.str() });
+envalid.cleanEnv(process.env, { ABSTRACT_TOKEN: envalid.str(), FIGMA_TOKEN: envalid.str() });
+
+const figmaNodeIds = ['1:22'];
```
jzempel, line 15: [nit] we should probably comment this array, like:

```js
const figmaNodeIds = [
  '1:22' /* home-hero-logo */
];
```

in order for code clarity going forward.

hzhu: Great idea, I pushed a commit up for this.

hzhu:
The `home-hero-logo` should match the one in production now.

jzempel (approved):

james-sann (approved):
`home-hero-logo` looks good to me!

PR #290, opened 2021-08-13
fix(tags): color example link to theme palette

jzempel:
## Description

Title sez it all.


smritimadan (approved):
Verified that the link works in staging environment.

hzhu (approved):

PR #295, opened 2021-09-09
chore(scripts): enhance deploy with added bandwidth checking

jzempel:
## Description

Verify available Netlify bandwidth prior to staging deploy.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,6 +5,7 @@
     "jsx": "preserve",
     "sourceMap": true,
     "strict": true,
+    "skipLibCheck": true,
     "moduleResolution": "node",
     "baseUrl": "./",
     "paths": {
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "jsx": "preserve",
     "sourceMap": true,
     "strict": true,
+    "skipLibCheck": true,
```
jzempel, line 8: Definitely less concerned about this here than on https://github.com/zendeskgarden/react-components/pull/1175.

utils/deploy.js
```diff
@@ -22,20 +22,30 @@ envalid.cleanEnv(process.env, {
...
+        const result = await garden.netlifyDeploy({
+          dir,
+          production,
+          message
+        });
+
+        return result;
+      };
+
+      url = await garden.githubDeploy({ command, production });
```
hzhu, line 43: Out of curiosity, if the GitHub deploy failed, would the error be bubbled up somewhere and handled?

hzhu (approved):

utils/deploy.js
jzempel: Yes. Note the caught, logged error below along with `process.exitCode = 1;` which would result in a CI fail.

hzhu: Thanks 👍🏽

PR #297, opened 2021-09-29
docs: add descriptions and notes to dropdown components

smritimadan:
## Description

This PR aims to restore and improve component documentation while following our content standards. Consideration was given to:
- providing information in a consistent, simple, clear and accessible way
- avoiding overloading with text or repetition
- using ways to make the content easier to scan quickly

Feedback will be highly appreciated in case I have missed historically important information :-).

## Detail

Addresses GARDEN-2344 and GARDEN-1375.

## Checklist


src/fragments/dropdowns.mdx
```diff
@@ -0,0 +1,15 @@
...
+for positioning. Use the `popperModifiers` prop of the Menu component to customize the placement, boundaries,
+overflows etc.
+
+**Dropdown State**: The [Downshift](https://github.com/downshift-js/downshift) `render-prop` library handles the
+keyboard and accessibility logic. Provide other customizations directly to the Downshift provider via the
+`downshiftProps` prop. See the [stateReducer](https://github.com/downshift-js/downshift#statereducer) pattern
+for an example of a common customization strategy.
+
+**Server Side Rendering**: Server side rendering might require configuring specific Downshift settings. The
+package mentioned re-exports the Downshift `resetIdCounter` utility, resetting the internal id counter which
```
mtomcal, line 14: Would you specifically name the package even if it's repetitive? I had a moment of confusion what package.

src/fragments/dropdowns.mdx
```diff
@@ -0,0 +1,15 @@
...
+is used to generate unique ids for Downshift.
```
mtomcal, line 15: Could we include `resetIdCounter` as part of the possible imports here:
Screen Shot 2021-09-29 at 11 45 46 AM

src/pages/components/autocomplete.mdx
```diff
@@ -98,48 +98,105 @@ import SizeCode from '!!raw-loader!../../examples/dropdowns/autocomplete/Size.ts
...
 ## API
 
+The Autocomplete component expects the following overall structure:
+
+```jsx isLight
+<Dropdown>
+  <Field>
+    <Label>
+    <Hint>
+    <Autocommplete>
```
mtomcal, line 115:
```suggestion
    <Autocomplete>
```

src/pages/components/multiselect.mdx
```diff
@@ -72,42 +72,98 @@ import MultiselectSizeCode from '!!raw-loader!../../examples/dropdowns/multisele
...
+- The dropdowns package does not supply filtering logic. The simplest way to perform filtering is to
+  control the `inputValue` and `onInputValueChange` props, and conditionally render Item components as
+  necessary.
+
+- Use input debouncing to delay Item rendering and prevent Downshift from reapplying accessibility
+  and interaction attributes.
+
+- Due to a known [issue](https://github.com/facebook/react/issues/11538#issuecomment-390386520) caused
+  by google translation extension, any conditionally rendered text node should be wrapped with a `span`
+  element.
```
mtomcal, line 114: Would this copy benefit from being in a fragment?

smritimadan: It might. I was having a hard time deciding where to draw the line for DRYing out repeated information, so I only did this for complete sections like the information on dropdowns. But I'm open to suggestions. Where would you draw the line?

mtomcal: This had 3 line breaks (3 bullets in this case) and constituted a subheading (Implementation notes) that was repeated. This triggered this thought to me. If it was one to three sentences and did not have it's own subheading I probably wouldn't split this out. I think it would be more maintainable as fragment imported to both pages.

src/pages/components/autocomplete.mdx
smritimadan: Fixed!

src/fragments/dropdowns.mdx
smritimadan: Updated!

src/fragments/dropdowns.mdx
smritimadan: It appears that this list is not meant to be an exhaustive list of all possible imports from the package, only an example? For instance, ItemMeta, MediaItem, MediaFigure, MediaBody etc are all valid subcomponents to use within an Autocomplete Menu, but aren't listed here. Could someone correct me if I'm wrong?

mtomcal: If that's the case, then we can forget about this comment. Unless @jzempel has other ideas we can let this point go.

mtomcal (approved):

src/pages/components/menu.mdx
```diff
@@ -141,22 +141,53 @@ import MenuSizeCode from '!!raw-loader!../../examples/dropdowns/menu/Size.tsx';
...
+```
+
+<!-- markdownlint-disable no-emphasis-as-heading -->
+
+**Implementation notes**
+
+<!-- markdownlint-enable no-emphasis-as-heading -->
+
+If loading is asynchronous, consider delaying Item rendering to prevent Downshift from reapplying accessibility and
+interaction attributes.
```
hzhu, line 164: I'm having a difficult time painting a picture of this scenario. How would a consumer delay `Item` rendering?

src/pages/components/menu.mdx
```diff
@@ -169,48 +200,76 @@ import MenuSizeCode from '!!raw-loader!../../examples/dropdowns/menu/Size.tsx';
 
 <Component components={props.data.mdx.components} componentName="item" />
 
+An Item is used for each dropdown option. It must be nested within a Menu component. It requires a `value` prop,
```
hzhu, line 203: What do you think about using the term "item" instead of "option"? 

Confusion regarding menu vs select dropdowns is common, so I think Garden can use words that better distinguish the two. Menu dropdown items have `role="menuitem"`. This is different for select dropdowns items which have `role="option"`.

hzhu (commented):
Thanks for putting this together! I'm curious where the restoration is coming from? Is it documentation that use to live in [v7](https://garden-v7.netlify.app/)? Or maybe somewhere in internal Google Docs?

src/fragments/dropdowns.mdx
jzempel: This list is intentionally populated via the frontmatter for the page, which should consist of the components most endemic to the main component of interest. Go to https://garden.zendesk.com/components/menu#configuration for the fuller listing more directly related to `Menu`.

This config is currently displaying as desired.

src/pages/components/menu.mdx
smritimadan: Great point. I had resorted to it because of how "An Item is used for each dropdown item" sounded. I'll try to rephrase it.

src/pages/components/multiselect.mdx
```diff
@@ -68,46 +68,89 @@ import MultiselectSizeCode from '!!raw-loader!../../examples/dropdowns/multisele
...
+
+<Dropdowns />
+
 <PropSheet components={props.data.mdx.components} componentName="dropdown" />
 
 ### Field
 
 <Component components={props.data.mdx.components} componentName="field" />
 
+A Field provides accessibility attributes to child fields. It must wrap the Autocomplete component.
```
jzempel, line 117: ~Autocomplete~ -> Multiselect

src/pages/components/select.mdx
```diff
@@ -105,46 +105,84 @@ import SelectValidationCode from '!!raw-loader!../../examples/dropdowns/select/V
...
+
+<Dropdowns />
+
 <PropSheet components={props.data.mdx.components} componentName="dropdown" />
 
 ### Field
 
 <Component components={props.data.mdx.components} componentName="field" />
 
+A Field provides accessibility attributes to child fields. It must wrap the Autocomplete component.
```
jzempel, line 151: *Select

jzempel (changes requested):

src/pages/components/menu.mdx
smritimadan: Fair point. I removed the part about an item being an item. 😅

src/pages/components/autocomplete.mdx
```diff
@@ -94,52 +94,96 @@ import SizeCode from '!!raw-loader!../../examples/dropdowns/autocomplete/Size.ts
...
+
 ## Configuration
 
 <Configuration reactPackage={props.data.mdx.package} components={props.data.mdx.components} />
 
 ## API
 
+The Autocomplete component expects the following overall structure:
+
+```jsx isLight
```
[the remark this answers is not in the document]
smritimadan: Thanks @jzempel, I'll remove `isLight` for this PR and get designers' opinion on this first. If the lighter treatment is desired I'll create a JIRA task to address this at `MarkdownProvider` level.

src/pages/components/autocomplete.mdx
```diff
@@ -94,52 +94,96 @@ import SizeCode from '!!raw-loader!../../examples/dropdowns/autocomplete/Size.ts
   </Do>
 </BestPractice>
 
+## Accessibility
+
+If no Label component is provided with the implementation, include an `aria-label` to provide an
+accessible name for the field.
+
```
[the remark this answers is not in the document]
smritimadan: I removed this section entirely so the addition of that section can be agreed upon first. 

My reason for adding this note was that even though we do have an example that shows the ideal way to hide the `Label` accessibly, it is currently entirely possible for teams today to just omit the `Label`. But thinking more about it, a better way might be to throw a dev-only console warning or something if a `Label` is not provided, instead of documenting a non-ideal approach like I was doing there.

src/pages/components/menu.mdx
```diff
@@ -169,48 +200,76 @@ import MenuSizeCode from '!!raw-loader!../../examples/dropdowns/menu/Size.tsx';
...
+which is provided to the `onSelect` callback.
+
+Item components can be wrapped and extended with styling libraries or other layout elements. All necessary
+information is provided through the Context API.
```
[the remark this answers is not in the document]
smritimadan: I wasn't entirely sure about this myself. It is likely an outdated note that is not very relevant now. Removed.

src/pages/components/menu.mdx
```diff
@@ -169,48 +200,76 @@ import MenuSizeCode from '!!raw-loader!../../examples/dropdowns/menu/Size.tsx';
...
+
 <PropSheet components={props.data.mdx.components} componentName="mediaFigure" />
 
 ### MediaItem
 
 <Component components={props.data.mdx.components} componentName="mediaItem" />
 
+The MediaItem component wraps the default Item component.
+
 <PropSheet components={props.data.mdx.components} componentName="mediaItem" />
```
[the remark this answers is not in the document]
smritimadan: Yes, I was wondering how `MediaItem` was different from `Item`. I can update the examples as part of this PR unless there are objections or any other reasons to wait.

src/pages/components/menu.mdx
```diff
@@ -169,48 +200,76 @@ import MenuSizeCode from '!!raw-loader!../../examples/dropdowns/menu/Size.tsx';
...
+A Menu manages placement, positioning and layout of the child Item components.
+
 <PropSheet components={props.data.mdx.components} componentName="menu" />
 
 ### NextItem
 
 <Component components={props.data.mdx.components} componentName="nextItem" />
 
+The NextItem component wraps the default Item component and displays the next icon to indicate the presence
+of a succeeding level. It enables additional RTL-aware keyboard navigation in nested menus.
```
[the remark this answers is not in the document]
smritimadan: Just verified this and yes, [it does have RTL-aware keyboard navigation](https://github.com/zendeskgarden/react-components/blob/71106c154e9d7d1b5926b3cc1cb2e0a3b3e174a6/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx#L100) built in 💯. I _could_ surface this comment in the section about the Dropdown component, but it only applies to this component in particular. 🤷‍♀️

src/fragments/dropdowns.mdx
```diff
@@ -0,0 +1,15 @@
...
+for positioning. Use the `popperModifiers` prop of the Menu component to customize the placement, boundaries,
+overflows etc.
+
+**Dropdown State**: The [Downshift](https://github.com/downshift-js/downshift) `render-prop` library handles the
+keyboard and accessibility logic. Provide other customizations directly to the Downshift provider via the
+`downshiftProps` prop. See the [stateReducer](https://github.com/downshift-js/downshift#statereducer) pattern
+for an example of a common customization strategy.
+
+**Server Side Rendering**: Server side rendering might require configuring specific Downshift settings. The
+dropdowns package re-exports Downshift's `resetIdCounter` utility, resetting the internal id counter which
```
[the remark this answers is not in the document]
smritimadan: Removed the first occurrence, changed the second one to ID. I'll bring this up with Diana once she has had a chance to gather some context around this epic.

src/pages/components/menu.mdx
jzempel: Plus 1: this phrasing re: reapplication of a11y attributes seen throughout feels off and is perhaps connected to an earlier component version that didn't handle conditional rendering as well as the current one does.

jzempel (changes requested):

src/pages/components/select.mdx
```diff
@@ -51,8 +51,8 @@ import SelectDefaultCode from '!!raw-loader!../../examples/dropdowns/select/Defa
 
 ### Disabled
 
-A disabled Select prevents user interaction. It doesn’t appear in the tab order, can’t receive focus, and may not read
-aloud by a screenreader.
+A disabled Select prevents user interaction. It doesn’t appear in the tab order, can’t receive focus, and may
+not read aloud by a screenreader.
```
jzempel, line 55: *not be read ...

src/pages/components/menu.mdx
```diff
@@ -223,6 +270,8 @@ import MenuSizeCode from '!!raw-loader!../../examples/dropdowns/menu/Size.tsx';
 
 <Component components={props.data.mdx.components} componentName="trigger" />
 
+A Trigger applies state and accessibility attributes to its children. It must be nested within a Dropdown component.
```
jzempel, line 273: Can we link `[Dropdown](#dropdown)` here?

src/pages/components/menu.mdx
smritimadan: Okay, this seems unnecessary in that case. Removed!

Francois-Esquire (approved):

src/pages/components/menu.mdx
```diff
@@ -199,18 +227,28 @@ import MenuSizeCode from '!!raw-loader!../../examples/dropdowns/menu/Size.tsx';
...
 <PropSheet components={props.data.mdx.components} componentName="menu" />
 
 ### NextItem
 
 <Component components={props.data.mdx.components} componentName="nextItem" />
 
+The NextItem component displays the next icon to indicate the presence of a succeeding level. It enables
+additional RTL-aware keyboard navigation in nested menus. The [Dropdown](#dropdown) component's `isOpen` and `onStateChange`
+props can be used to conditionally handle the open/close state based on selection.
+
```
hzhu, line 241: Fleshing this part out is great! Many consumers mistakenly nest items in `Select` components. Providing some content around `NextItem` helps emphasize that nesting structure belongs in `Menu` rather than `Select`.

src/pages/components/menu.mdx
```diff
@@ -199,18 +227,28 @@ import MenuSizeCode from '!!raw-loader!../../examples/dropdowns/menu/Size.tsx';
...
 
+A Menu manages placement, positioning and layout of the child [Item](#item) components.
+
 <PropSheet components={props.data.mdx.components} componentName="menu" />
 
 ### NextItem
 
 <Component components={props.data.mdx.components} componentName="nextItem" />
 
+The NextItem component displays the next icon to indicate the presence of a succeeding level. It enables
```
diana-oum, line 238: indicate --> show (simplifying the verb!)

src/pages/components/menu.mdx
```diff
@@ -199,18 +227,28 @@ import MenuSizeCode from '!!raw-loader!../../examples/dropdowns/menu/Size.tsx';
...
+additional RTL-aware keyboard navigation in nested menus. The [Dropdown](#dropdown) component's `isOpen` and `onStateChange`
+props can be used to conditionally handle the open/close state based on selection.
+
 <PropSheet components={props.data.mdx.components} componentName="nextItem" />
 
 ### PreviousItem
 
 <Component components={props.data.mdx.components} componentName="previousItem" />
 
+The PreviousItem component displays the previous icon to indicate the presence of a preceding level. It enables
```
diana-oum, line 248: indicate --> show

diana-oum (approved):
Nice work Smriti. This review didn't include your option 1/option 2. I accidentally commented on that outside of the review (oops, sorry). After that and these two verb changes, it's good to go!

smritimadan:
> Looks like [some spacing weirdness](https://share.getcloudapp.com/bLudDool) btwn the filtering notes and the “Autocomplete” H3. Same issue on the Multiselect page. I think it’s correct mathmatically, but it looks much smaller visually than some of the other spacings around H3s. It’s probably fine as is, so I’m approving, but please do double check to make sure!
> 
> I think it’s just part of the stacking weirdnessit should match the spacing seen elsewhere (feels too small currently)

Good catch! Looks like this spacing discrepancy is coming from the variations in the margin on the elements immediately preceding the H3s (screen grabs below show the margins on a preceding unordered list, table and paragraph.) I _could_ apply a custom theme object to override the margin in this individual case, but I feel that this is something that we might need to address holistically at the `MarkdownProvider` level while making sure we don't cause regressions in other places. Not sure if other devs have any ideas on this.

Screen Shot 2021-10-15 at 4 32 38 PM
Screen Shot 2021-10-15 at 4 35 31 PM
Screen Shot 2021-10-15 at 4 36 01 PM

allisonacs:
> Good catch! Looks like this spacing discrepancy is coming from the variations in the margin on the elements immediately preceding the H3s (screen grabs below show the margins on a preceding unordered list, table and paragraph.) I _could_ apply a custom theme object to override the margin in this individual case, but I feel that this is something that we might want to address holistically at the `MarkdownProvider` level while making sure we don't cause regressions in other places. Not sure if other devs have any ideas on this.

I thought might be the case. OK, ship this as is. Please file both design and eng tickets to address spacing for H3s on the site. I looked at the original files, and the H3s were implemented as spec’d, so we’ll need designer eyes to determine top margin for us.

smritimadan:
> I thought might be the case. OK, ship this as is. Please file both design and eng tickets to address spacing for H3s on the site. I looked at the original files, and the H3s were implemented as spec’d, so we’ll need designer eyes to determine top margin for us.

Sounds good, I'll create tickets. Thanks!

PR #314, opened 2021-10-15
feat(fileupload): add new `FileList` component documentation

jzempel:
## Description

- Lock existing `FileUpload` examples down with associated `FileList` content
- Add new `File` example (@lucijanblagonic I think we were able to statically demonstrate the files you hoped for)
- Add associated Configuration & API docs

## Checklist

### DIFF
--- a/src/examples/forms/file-upload/FileUpload.tsx
+++ b/src/examples/forms/file-upload/FileUpload.tsx
@@ -5,15 +5,93 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React from 'react';
+import React, { useCallback, useState, memo } from 'react';
+import styled from 'styled-components';
 import { useDropzone } from 'react-dropzone';
-import { Field, Label, Hint, Input, FileUpload, Message } from '@zendeskgarden/react-forms';
+import { KEY_CODES } from '@zendeskgarden/container-utilities';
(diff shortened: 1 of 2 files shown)


jzempel:
@lucijanblagonic the `isCompact` applies to both the `FileUpload` and `FileList`. The issue you are seeing here is due to a recent bug we uncovered with `styled-components` related to instance specificity. (See https://github.com/zendeskgarden/react-components/pull/1225) Once those bug fixes on `react-components` are published and make their way over here, the compact `Label` will render 4px away from the upload (vs. the current 8px). In lieu of that bug fix, I think the existing copy makes sense.

src/examples/forms/file-upload/FileUpload.tsx
```diff
@@ -36,6 +112,15 @@ const Example = () => {
             )}
             <Input {...getInputProps()} />
           </FileUpload>
+          {files.length === 0 ? (
+            <Message>Acceptable formats are JPG, PNG, and GIF</Message>
+          ) : (
+            <StyledFileList>
+              {files.map((file, index) => (
+                <FileItem key={file} name={file} onRemove={() => removeFile(index)} />
```
hzhu, line 120: Alternatively, the `onRemove` could pass the `index` argument implicitly.

PR #324, opened 2021-12-03
feat(combobox): add documentation page for component

Francois-Esquire:
## Description

Adding documentation page for combobox to the website.

## Detail

There are a few things that still need to be addressed:

- [x] add fuzzy search to examples and update search items
- [x] the do/don't example image is not loading properly

## Checklist


src/examples/dropdowns/combobox/Default.tsx
```diff
@@ -0,0 +1,84 @@
...
+      const matchedOptions = options.filter(
+        option => option.trim().toLowerCase().indexOf(value.trim().toLowerCase()) !== -1
+      );
+
+      setMatchingOptions(matchedOptions);
+    }, 300)
+  );
+
+  useEffect(() => {
+    filterMatchingOptionsRef.current(inputValue);
```
mtomcal, line 50: Would clarify using a Ref for storing a function versus using a `useCallback` style function?

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,204 @@
+---
+title: Combobox
+description: Combobox is an input plus a list box that appears beneath it. It is used for search and filtering.
```
mtomcal, line 3: Screen Shot 2021-12-03 at 12 01 50 PM
Could we use active voice?

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,204 @@
...
+import ComboboxMedia from '../../examples/dropdowns/combobox/Media.tsx';
+import ComboboxMediaCode from '!!raw-loader!../../examples/dropdowns/combobox/Media.tsx';
+
+<CodeExample code={ComboboxMediaCode}>
+  <ComboboxMedia />
+</CodeExample>
+
+### Selection Logic
+
+The combobox can can be customized to include selection logic that highlights the first relevant
```
mtomcal, line 68: `can can`

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,204 @@
...
+Nest a Hint within a [Field](#field) component.
+
+<PropSheet components={props.data.mdx.components} componentName="hint" />
+
+### Item
+
+<Component components={props.data.mdx.components} componentName="item" />
+
+Nest an Item within a [Menu](#menu) component. It requires a `value` prop, which is provided to
+the `onSelect` callback.
```
mtomcal, line 162: Passive voice to active voice?

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,204 @@
...
+    <ul>
+      <li>To filter down a list of options</li>
+      <li>To quickly find a known option</li>
+      <li>As a search field</li>
+    </ul>
+  </Use>
+  <Misuse>
+    <ul>
+      <li>For a searchable dropdown, use Autocomplete instead</li>
+      <li>To make more than one selection, use Multiselect instead</li>
```
jzempel, line 26: Please link "Autocomplete" and "Multiselect" similar to https://garden.zendesk.com/components/autocomplete

jzempel (changes requested):
As discussed offline, there's a UI jitter on first component focus. Also, can we get the filter & sorting to occur before the initial menu display? The timing on the visual re-sort after initial expansion feels a bit off.

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,204 @@
...
+
+## How to use it well
+
+<BestPractice>
+  <Do imageSource={props.data.comboboxHighlightDo.childFile.childImageSharp.gatsbyImageData}>
+    <p>Use balanced and appropriate typographic hierarchy.</p>
+  </Do>
+  <Dont imageSource={props.data.comboboxSearchableDont.childFile.childImageSharp.gatsbyImageData}>
+    <p>Don’t mix up typographic hierarchy when customizing item content.</p>
+  </Dont>
```
hzhu, line 98: These best practices don't seem to reflect the Figma asset.

src/pages/components/combobox.mdx
Francois-Esquire: @diana-oum what are your thoughts here?

src/examples/dropdowns/combobox/Default.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <Row justifyContent="center">
+      <Col sm={5}>
+        <Dropdown
+          inputValue={inputValue}
+          selectedItem={selectedItem}
+          onSelect={item => setSelectedItem(item)}
+          onInputValueChange={value => setInputValue(value)}
+          downshiftProps={{ defaultHighlightedIndex: -1 }}
```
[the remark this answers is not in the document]
Francois-Esquire: Good catch! Will remove these and duplicates.

src/pages/components/combobox.mdx
Francois-Esquire: Updated! Thanks for spotting this

src/pages/components/combobox.mdx
Francois-Esquire: Good catch, updated!

src/examples/dropdowns/combobox/Default.tsx
Francois-Esquire: Good call, a `useCallback` would work perfectly with the `debounce` invocation directly, however the linter gave me some trouble, so I opted with a `useMemo` instead here.

src/pages/components/combobox.mdx
Francois-Esquire: updated to:
```
Nest an Item within a [Menu](#menu) component. It requires a `value` prop, which it provides to
the `onSelect` callback.
```

src/pages/components/combobox.mdx
Francois-Esquire: Yes an image seems to be displayed incorrectly, if that is what you mean - or did you have something else in mind?

Francois-Esquire:
> As discussed offline, there's a UI jitter on first component focus. Also, can we get the filter & sorting to occur before the initial menu display? The timing on the visual re-sort after initial expansion feels a bit off.

@jzempel The filter & sorting issue should be gone now 🤞  looking more into the jitter and why that is happening..

src/pages/components/combobox.mdx
hzhu: Seems like the copy for "Do This" and "Not This" might have been left over from a copy paste of `timeline.tsx`. Typographic hierarchy doesn't seem to be related to combo box:

Do This:

> When using Combobox as a search input, automatically highlight the first suggestion in the list box and allow the user to hit the Enter key to submit the search.

Not This:

> Use an Autocomplete component when creating a searchable dropdown.

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,231 @@
...
+import ComboboxSelection from '../../examples/dropdowns/combobox/Selection.tsx';
+import ComboboxSelectionCode from '!!raw-loader!../../examples/dropdowns/combobox/Selection.tsx';
+
+<CodeExample code={ComboboxSelectionCode}>
+  <ComboboxSelection />
+</CodeExample>
+
+## Usage
+
+<BestPractice rows>
```
hzhu, line 87: Is the `rows` prop used for something? I wasn't able to find a `rows` prop usage in `<BestPractice>`.

package.json
```diff
@@ -78,6 +78,7 @@
     "eslint-plugin-notice": "0.9.10",
     "eslint-plugin-react": "7.26.1",
     "eslint-plugin-react-hooks": "4.2.0",
+    "fuzzyjs": "5.0.1",
```
hzhu, line 81: What are the reasons for adding `fuzzyjs` and is regular filtering insufficient? Is it recommended that consumers also use this dependency when using `Combobox`?

src/pages/components/combobox.mdx
diana-oum: @Francois-Esquire - "It searches and filters down options for the user."

I also think "Input" is actually the name of a component. Shouldn't that be linked?

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,204 @@
...
+import ComboboxDefault from '../../examples/dropdowns/combobox/Default.tsx';
+import ComboboxDefaultCode from '!!raw-loader!../../examples/dropdowns/combobox/Default.tsx';
+
+<CodeExample code={ComboboxDefaultCode}>
+  <ComboboxDefault />
+</CodeExample>
+
+### Hidden Label
+
+The label can be hidden as needed.
```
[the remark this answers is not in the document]
diana-oum: Great catch. Let's keep it simple and concise and not include "as needed". I assume that we assume people reading the documentation will make that call and we're simply explaining that it CAN be hidden  cc: @bethanybeech

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,204 @@
...
+import ComboboxHidden from '../../examples/dropdowns/combobox/Hidden.tsx';
+import ComboboxHiddenCode from '!!raw-loader!../../examples/dropdowns/combobox/Hidden.tsx';
+
+<CodeExample code={ComboboxHiddenCode}>
+  <ComboboxHidden />
+</CodeExample>
+
+### Media
+
+You can place the magnifying glass icon at the start of the input when using the Combobox for search.
```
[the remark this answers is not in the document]
diana-oum: I don't have context on Combobox unfortunately. This was done before I joined and I just reviewed it from a legibility standpoint. Will let @bethanybeech chime in regarding the icon. 

But I agree Henry if there isn't a non-search use case documented here, it sounds weird for the documentation to preface things like using phrases "for search" and we should omit them.

src/pages/components/combobox.mdx
Francois-Esquire: Ahh I misunderstood what was meant by assets (thinking of images). Good catch on the copy, that has been updated!

src/pages/components/combobox.mdx
Francois-Esquire: It was a new addition added to support Do/Don't examples that span the entire row (per example) from the Figma design, instead of structuring the content in columns (which is the default). Could this be accomplished in a different way you think?

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,231 @@
...
+import ComboboxHiddenCode from '!!raw-loader!../../examples/dropdowns/combobox/Hidden.tsx';
+
+<CodeExample code={ComboboxHiddenCode}>
+  <ComboboxHidden />
+</CodeExample>
+
+### Selection Logic
+
+The combobox can be customized to include selection logic that highlights the first relevant
+result as the user types, allowing the user to hit enter to fully select the item. If this logic
```
diana-oum, line 75: I'm noticing this now: "hit enter" sounds really colloquial in (American?) English. 
Let's change this to "...allowing the user to `press Enter to select the item`". 

I also want to remove "fully" because I don't think it's adding anything to the sentence in terms of understanding

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,231 @@
...
+      search.
+    </p>
+  </Do>
+  <Do
+    imageIsFreeForm
+    imageSource={props.data.comboboxHighlightDo.childFile.childImageSharp.gatsbyImageData}
+  >
+    <p>
+      When using Combobox as a search input, automatically highlight the first suggestion in the
+      list box and allow the user to hit the Enter key to submit the search.
```
diana-oum, line 103: Same change here! "hit" --> "press"

src/pages/components/combobox.mdx
hzhu: I see, thanks for clarifying. LGTM.

src/examples/dropdowns/combobox/Default.tsx
```diff
@@ -0,0 +1,91 @@
...
+              <Item disabled>No matches found</Item>
+            )}
+          </Menu>
+        </Dropdown>
+      </Col>
+    </Row>
+  );
+};
+
+export default Example;
```
hzhu, line 91: I noticed that when a user removes the characters from the text box after a search, the list box with all the options flash for a split second. If this is a bug, is there a way to fix it?

mtomcal: It's like it needs an extra ms of `debounce` maybe 🤔

package.json
jzempel: @Francois-Esquire can we revert to what you had before? The `fuzzyjs` is not available to codesandbox, and we want to keep the raw demo library set to the bare minimum.

src/pages/components/combobox.mdx
Francois-Esquire: Thank you @diana-oum ! That works, will update.

src/pages/components/combobox.mdx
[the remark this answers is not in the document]
Francois-Esquire: Good catch @hzhu and thanks for clarifying @diana-oum, I have updated the copy.

package.json
Francois-Esquire: @jzempel no problem, I've updated and reverted to how it was previously.

@hzhu to give some context, @jzempel and I were discussing fuzzy search and after some research into this, I figured I'd rely on a pre-existing solution with `fuzzyjs`.

src/pages/components/combobox.mdx
Francois-Esquire: Updated!

src/pages/components/combobox.mdx
Francois-Esquire: Good catch, updated!

src/examples/dropdowns/combobox/Default.tsx
Francois-Esquire: I think I introduced this behavior while adding the option `{ leading: true }` to the `debounce` function. I removed it and added better indication of when the Combobox is searching.

src/pages/components/combobox.mdx
Francois-Esquire: Screen Shot 2021-12-13 at 10 02 33 PM

Super odd, not updated on the PR, yet locally when pulled down, it reflects accurately to what was updated..

src/examples/dropdowns/combobox/Default.tsx
```diff
@@ -0,0 +1,103 @@
...
+          inputValue={inputValue}
+          selectedItem={selectedItem}
+          onSelect={item => setSelectedItem(item)}
+          onInputValueChange={value => setInputValue(value)}
+        >
+          <Field>
+            <Label>Choose a vegetable</Label>
+            <Combobox
+              start={<SearchIcon />}
+              end={isLoading ? <Dots delayMS={0} /> : null}
```
jzempel, line 80: I'm still seeing the jitter on hover after initial load. Have you looked into removing these lines? The component was never tested in conjunction with the `<Dots>` loader. In general, the intent of these start/end slots is not for containing loader assets, but rather for icons.  We don't currently have examples on the site, but in the past the input-loader pattern has been expressed in a different manner altogether – see the "Loading items..." treatment under https://garden.zendesk.com/components/multiselect#default. This would also mitigate the false "No matches found" message that briefly appears prior to a successful search.

I would recommend removing this `end` throughout and seeing if that alleviates the jitter-on-hover.

src/examples/dropdowns/combobox/Default.tsx
```diff
@@ -0,0 +1,103 @@
...
+            options.filter(
+              option => option.trim().toLowerCase().indexOf(value.trim().toLowerCase()) !== -1
+            ) || []
+          );
+        } else {
+          setMatchingOptions([]);
+        }
+        setLoading(false);
+      }, 750),
+    [setMatchingOptions]
```
hzhu, line 56: Alternatively, can remove `setMatchingOptions` from dependency array since it's a set state from React and will not be changing each render.

src/examples/dropdowns/combobox/Default.tsx
```diff
@@ -0,0 +1,103 @@
...
+  );
+
+  useEffect(() => {
+    if (inputValue) {
+      setLoading(true);
+    }
+
+    filterMatchingOptions(inputValue);
+  }, [filterMatchingOptions, inputValue]);
```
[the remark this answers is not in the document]
Francois-Esquire: Ahh, thanks for pointing that out.. at some point I had the `debounce` function initialized outside the component, will revert to that.

src/examples/dropdowns/combobox/Default.tsx
Francois-Esquire: I see the jitter on hover now.. Hmm, both the loaders and icons are technically SVG elements so I don't immediately suspect the `Dots` component, I'll test it out without the `Dots` component and see it it helps.

src/examples/dropdowns/combobox/Default.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  return (
+    <Row justifyContent="center">
+      <Col sm={5}>
+        <Dropdown
+          inputValue={inputValue}
+          selectedItem={selectedItem}
+          onSelect={item => setSelectedItem(item)}
```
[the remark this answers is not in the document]
Francois-Esquire: Good catch, will get this fixed.

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,233 @@
...
+      <li>To filter down a list of options</li>
+      <li>To quickly find a known option</li>
+      <li>As a search field</li>
+    </ul>
+  </Use>
+  <Misuse>
+    <ul>
+      <li>
+        <Markdown>
+          For a searchable dropdown, use [Autocomplete](/components/autocomplete) instead
```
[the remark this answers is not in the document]
Francois-Esquire: Well it would be considered a dropdown and it is searchable.. unlike the `Select`, `Multiselect` and `Autocomplete`, the Combobox doesn't have a chevron to drop down a menu with this component. The combobox is amorphous since it technically can be a `Multiselect` and `Autocomplete` [according to the spec](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html).

Do you think we should change this?

src/examples/dropdowns/combobox/Default.tsx
Francois-Esquire: You are right, will update and remove.

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,233 @@
...
+  - dropdowns/src/elements/Menu/Items/Item.tsx
+  - dropdowns/src/elements/Fields/Label.tsx
+  - dropdowns/src/elements/Menu/Menu.tsx
+---
+
+<Usage>
+  <Use>
+    <ul>
+      <li>To filter down a list of options</li>
+      <li>To quickly find a known option</li>
```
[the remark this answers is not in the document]
Francois-Esquire: Good question. How I understand it is that the user behavior is to selectively display from a given set of results. For example, searching for phone number area codes or countries would be a good use of the Combobox.

jzempel (changes requested):
Keyboard selection is not working. Also, the menu is still showing "No matches found" when it should be showing "Loading items...". Can we just make this demo work almost exactly like [Multiselect](https://garden.zendesk.com/components/multiselect) except without a) any selection and b) menu expansion on focus?

When we met with the teams who originally asked for the `Combobox`, the request was to basically supply something with behavior similar to a Google search. So this component page should demonstrate behavior very close to that basic experience (without the complexity of a backend and fuzzy search). It should work very much like what we see [in Storybook](https://61d5b1f8141e3173753a1140--zendeskgarden-react-components.netlify.app/?path=/story/packages-dropdowns-combobox--combobox) with the addition of rudimentary filtering to narrow down the menu items based on input. No checkmark selections.

Also, I think I've isolated the UI jitter that is seen on load/hover. It looks like we have another case of a `styled-components` v5 precedence boost issue 😬 where the `Label` > `Input` initially styles incorrectly compact and then default spacing is overridden on hover. We'll need to address this problem in the `react-components` view component code.

src/pages/components/combobox.mdx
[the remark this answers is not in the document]
hzhu: Ah gotcha, thanks for clarifying!

src/pages/components/combobox.mdx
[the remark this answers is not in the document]
hzhu: I see. I don't have a suggestion at this time. But maybe we can revisit this if consumers need clarification.

src/examples/dropdowns/combobox/Default.tsx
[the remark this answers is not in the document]
hzhu: Oh I think it's fine inside the component like you have it. Cleaning up after an effect would resolve issues relating to potential memory leaks, expected behaviors, etc.

```jsx
useEffect(() => {
  return {
    debouncedFn.cancel();
  }
}, [])
```

Francois-Esquire:
@jzempel thank you for sharing all this! I've updated the PR to address all the points mentioned (except the UI jitter, which we can track and tackle at another time).

src/examples/dropdowns/combobox/Default.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  return (
+    <Row justifyContent="center">
+      <Col sm={5}>
+        <Dropdown
+          inputValue={inputValue}
+          onInputValueChange={value => setInputValue(value)}
+          selectedItem={selectedItem}
+          onSelect={item => {
+            setInputValue(item);
```
Francois-Esquire, line 87: I had to set the input value on select since it was not reflecting on the input without it. I had assumed that `selectedItem` would take care of this for the Combobox.

hzhu (commented):
I think JZ had [mentioned](https://github.com/zendeskgarden/website/pull/324#pullrequestreview-844901374) that there should not be a SVG check icon next to the selected option.

Francois-Esquire:
> I think JZ had [mentioned](https://github.com/zendeskgarden/website/pull/324#pullrequestreview-844901374) that there should not be a SVG check icon next to the selected option.

@hzhu good catcht. Updated!

hzhu (approved):
LGTM, after the comments from other Gardners are addressed. It looks like the jitter will be fixed in a follow up PR?

src/pages/components/combobox.mdx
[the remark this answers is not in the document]
jzempel: In a reconsidered world, `Combobox` would've been the `Dropdown` baseline and `Autocomplete`, `Multiselect`, and (maybe) `Select` would've inherited from that flexible baseline. With this inheritance in mind, all three of `Combobox` (a no-select), `Autocomplete` (a single select), and `Multiselect` (erm multi-select) are searchable. And all three drop down a menu. So I'm kind of tripping on the wording here as well.

jzempel (approved):

src/pages/components/combobox.mdx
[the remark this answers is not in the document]
diana-oum: Mike is correct that the combobox is not a dropdown because it does not have a chevron to show a menu of items. Looking at the components within the Dropdown group today, they all have this. Combobox does not.

Therefore it's an issue to have Combobox in the Dropdown group--it's incorrect and confusing. For now, I think the simplest path is to have Combobox not be under the Dropdowns group and on a separate page of its own. That way it makes sense while we don't have to overhaul the entire information architecture. Thoughts?

Sorry I'm giving this kind of feedback now. I was not aware of where we were placing this component page. The rest of the content looks good. Thanks Mike!

cc: @bethanybeech

diana-oum (changes requested):

bethanybeech:
The page looks great! I agree with @diana-oum that placing this within Dropdowns does feel odd. I understand that it is built off of a dropdown component, but functionally it behaves differently, which could be confusing to non-technical users. If it's possible to place it outside of the Dropdowns section, it could help with that confusion.

mtomcal:
Per @diana-oum and @bethanybeech's comments on new page outside of Dropdowns for Combobox. Would this be a breaking change (v9 roadmap item) in that we would also need to alter which package (`@zendeskgarden/react-dropdowns`) the Combobox belongs to if we change the website IA?

src/pages/components/combobox.mdx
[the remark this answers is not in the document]
hzhu: Does the absence of a chevron icon indicate `Combobox` is not a dropdown? I don't think the word has been defined in Garden, so I'd be curious of how the team defines this. My definition is that a dropdown is a widget that opens or expands when it is selected, revealing a list of items. Although Garden's combo box does not have an icon, it is optional according to ARIA:

> Many implementations also include a third optional element -- a graphical button adjacent to the textbox, indicating the availability of the popup.

Source: https://www.w3.org/TR/wai-aria-practices/#combobox

In Garden, there's a `Dropdown` component that manages the behavior and state of components that belong to the "dropdown family" ([select](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html), [combobox](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html), [menu button](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html)). Note the similarities of each:

```jsx
// dropdown manages state for `Combobox`
<Dropdown>
  <Field>
    <Label>Choose a vegetable</Label>
    <Combobox placeholder="Search vegetables" />
  </Field>
  <Menu>
    <Item>Spinach</Item>
    <Item>Kale</Item>
    <Item>Lettuce</Item>
  </Menu>
</Dropdown>
```

```jsx
// dropdown manages state for `Autocomplete`
<Dropdown>
  <Field>
    <Label>Choose a vegetable</Label>
    <Autocomplete>Spinach</Autocomplete>
  </Field>
  <Menu>
    <Item>Spinach</Item>
    <Item>Kale</Item>
    <Item>Lettuce</Item>
  </Menu>
</Dropdown>
```

```jsx
// dropdown manages state for `Select`
<Dropdown>
  <Field>
    <Label>Choose a vegetable</Label>
    <Select>Spinach</Select>
  </Field>
  <Menu>
    <Item>Spinach</Item>
    <Item>Kale</Item>
    <Item>Lettuce</Item>
  </Menu>
</Dropdown>
```

```jsx
// dropdown manages state for Menu`
<Dropdown>
  <Trigger>
    <Button>
      Choose vegetable
      <Button.EndIcon isRotated={rotated}>
        <ChevronIcon />
      </Button.EndIcon>
    </Button>
  </Trigger>
  <Menu>
    <Item>Spinach</Item>
    <Item>Kale</Item>
    <Item>Lettuce</Item>
  </Menu>
</Dropdown>
```

These reasons lead me to believe that `Combobox` is a dropdown and belongs in the `react-dropdowns` package. 

It seems like components on the website don't necessarily have to be grouped by their packages. For example, `Stepper` and `Timeline` are part of the accordion 🪗 package, but are visible at the root level of the side panel. My understanding is that this setup highlights the `Stepper` and `Timeline` as significant in itself. I think `Combobox` can be visible at the root level if it is significant in itself, or makes things more clear for consumers.

tldr: I think `Combobox` is a dropdown but can probably still live at the root level (if it helps make things more clear).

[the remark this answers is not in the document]
diana-oum: >  It seems like components on the website don't necessarily have to be grouped by their packages. For example, Stepper and Timeline are part of the accordion 🪗 package, but are visible at the root level of the side panel. My understanding is that this setup highlights the Stepper and Timeline as significant in itself. I think Combobox can be visible at the root level if it is significant in itself, or makes things more clear for consumers.

Well said @hzhu I think it would be worthwhile for Combobox to also be at the root for clarity.

To answer your question, my understanding of what a dropdown is informed by other design systems which is why I think the chevron is a distinction, not saying that's the correct definition or that any of these other design systems have it right:
https://atlassian.design/components/dropdown-menu/examples
https://clarity.design/angular-components/dropdown/
https://material.io/components/menus#dropdown-menu
https://design.cms.gov/components/dropdown/

Thinking on this more, I believe it's less about the chevron to your point and ARIA, and more about the interaction. 

The difference between combobox and the other "Dropdown" components for Garden is very small as a reader of the documentation so that is a challenge here. But a combobox has a listbox that appears as you type in the input whereas autocomplete has dropdown that appears when the user clicks into the input. User has to click to get into the box for components Menu, Select, and Multiselect.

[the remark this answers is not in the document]
Francois-Esquire: If it makes things more clear @diana-oum , let's discuss this deeper because I do have some concerns with what @hzhu & @jzempel brought up. Does dropdown mean having a chevron to fit the bill or is it any component that drops a list of items? Another concern of mine is convention, and changing it would contrast how we've structured the project, which is something to consider..

I don't mind making the change if we agree it would clarify things and makes sense, and we've done it in the past as @hzhu pointed out. Ultimately, how should we define Dropdown?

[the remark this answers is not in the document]
Francois-Esquire: Another thought I had goes back to the phrase "searchable dropdown". Do you think we should rephrase it more specifically? Maybe "searchable and selectable dropdown"?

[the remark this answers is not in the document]
diana-oum: @Francois-Esquire I agree, this is a conversation that really should have required us earlier in the process for all of us to come together and chat about. It's a large decision for sure. But since we're trying to wrap up Q4 things ASAP, let's just keep this learning in mind when working on future component pages. 

>  Another thought I had goes back to the phrase "searchable dropdown". Do you think we should rephrase it more specifically? Maybe "searchable and selectable dropdown" could remove some confusion?

That is something to consider and one that I'm gonna try to push for instead to keep the convention we have now to get us moving forward. Let me brainstorm with Bethany as she was the driver of the documentation.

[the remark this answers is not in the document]
diana-oum: @Francois-Esquire - I jammed with Bethany on this just now. Let's replace that first bullet in the Not for this section with this sentence: "To see a list of options in the dropdown without entering text in the input field, use Autocomplete" We'll maintain the concept model that's already established while trying to be clear as we can.

Based on this thread and our own confusion, there are terminology issues here. But that needs to be addressed in a separate initiative where we all can be involved from the beginning! Let's also learn from our consumers like Henry already mentioned. Appreciate everyone's input here. Thanks!

hzhu:
@mtomcal I think the package structure can remain the same. It seems like the website IA changes and package structure can be mutually exclusive.

src/pages/components/combobox.mdx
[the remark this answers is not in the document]
Francois-Esquire: @diana-oum let's take this offline since there are other things to consider.

src/pages/components/combobox.mdx
```diff
@@ -0,0 +1,233 @@
...
+      list box and allow the user to press the Enter key to submit the search.
+    </p>
+  </Do>
+  <Dont
+    imageIsFreeForm
+    imageSource={props.data.comboboxSearchableDont.childFile.childImageSharp.gatsbyImageData}
+  >
+    <p>
+      <Markdown>
+        Use an [Autocomplete](/components/autocomplete) component when creating a searchable
```
diana-oum, line 112: @Francois-Esquire  Replacing the problematic "searchable dropdown" phrase and changing the content to this:

"Use an Autocomplete component when creating a dropdown that shows a list of options without needing to enter text in the input field."

Francois-Esquire: For sure, just updated.

diana-oum (approved):

PR #342, opened 2022-02-02
chore: generate subcomponent API directly from source code

jzempel:
## Description

Update to react-components v8.48.0 with refactored subcomponents which provide naming and data that can be parsed via docgen. This PR introduces new `subcomponents:` page frontmatter that can be used to target subcomponent source without adding these to the page's `import` configuration. The result is that each Garden subcomponent can receive `<Component>` and `<PropSheet>` components for information display, just like top-level components. Most importantly, we no longer need to maintain hand-rolled `components.ts` constructs for subcomponents, allowing the data to be derived directly from source code.

## Detail

Please note that the addition of `<PropSheet>` components will only display when props are present, making the page data more future-proof if/when props are added to any subcomponents.


mtomcal (approved):
This looks like a welcome addition 👍

Francois-Esquire (approved):

Francois-Esquire:
This PR makes our documentation much more resilient and maintenance-free when it comes to the subcomponent API surface, thanks @jzempel 🙌

hzhu (approved):

PR #348, opened 2022-02-10
feat(icons): New Icon Overview & Icon Search

exelarios:
## Description
This pull request introduces the level 1 stage of icon documentation implementation, which consists of: 
- Icon Overview Page
- Icon Library w/ search & collapse.

## Detail
Things to still consider:
- Should the icon filtering be smooth with animations?
- How accessible is the gradient bottom of the icons?
- Does the icon search need a way to re-collapse?

**Overview & Navbar Structure**
image

**Icon Library**
image

## Checklist

### DIFF
--- a/src/examples/design/icons/SvgSearch.tsx
+++ b/src/examples/design/icons/SvgSearch.tsx
@@ -142,7 +142,7 @@ export const SvgSearch: React.FC<ISvgSearchProps> = ({ data, searchEnabled }) =>
                 margin-bottom: ${p => p.theme.space.lg};
               `}
             >
-              <Label>Filter icons</Label>
+              <Label>Search icons</Label>
               <MediaInput start={<SearchStroke />} onChange={onInputChange} />
             </Field>
           </Col>


src/examples/design/icons/IconShowcase.tsx
```diff
@@ -0,0 +1,41 @@
...
+import React from 'react';
+import { Grid, Row, Col } from '@zendeskgarden/react-grid';
+import { PALETTE } from '@zendeskgarden/react-theming';
+import { ReactComponent as LeafIcon } from '@zendeskgarden/svg-icons/src/16/leaf-stroke.svg';
+import { ReactComponent as LightBulbIcon } from '@zendeskgarden/svg-icons/src/16/lightbulb-stroke.svg';
+import { ReactComponent as LightningBoltIcon } from '@zendeskgarden/svg-icons/src/16/lightning-bolt-stroke.svg';
+import { ReactComponent as LocationIcon } from '@zendeskgarden/svg-icons/src/16/location-stroke.svg';
+import { ReactComponent as SmileyIcon } from '@zendeskgarden/svg-icons/src/16/smiley-stroke.svg';
+
+const ICON_FILL = PALETTE.grey[600];
```
hzhu, line 17: Should this leverage the `theme` value through styled-components? Consumers of the design system should not access a color by pulling directly from `PALETTE`.

src/examples/design/icons/IconStyles.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <Grid>
+      <Row justifyContent="center">
+        <StyledCol size={3}>
+          <LeafStrokeIcon />
+          <MD>Stroke style (default)</MD>
+        </StyledCol>
+        <StyledCol size={3}>
+          <LeafFillIcon />
+          <MD>Filled Styled</MD>
```
hzhu, line 31: These words read a bit strange because [stroke](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke) and [fill](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill) are words used to describe attributes of SVG and words aren't typically used as past tense verbs. Also, it reads strange because one "style" is a noun, while the other is a past tense verb, "styled". I think _"Stroke (default)"_ and _"Fill"_ reads more clean than _"Stroke style (default)"_ and _"Filled Styled"_.

src/components/MarkdownProvider/components/BestPractice.tsx
```diff
@@ -118,6 +123,7 @@ interface ISectionProps extends ICaptionProps {
   imageHeight?: number;
   imageWidth?: number;
   imageIsSquare?: boolean;
+  imageBackgroundColor?: string;
```
hzhu, line 126: The image should be sized accordingly to fill in the space. I think design should update the image so that it fits correctly.

Note on props: I think adding new props should be considered carefully. If a new prop is added for each new requirement, it could lead to components having dozens of props with take away rendering control from the consumer. IMO, props should only be added if its critical part of component implementation — otherwise build the component such that consumers can compose behavior + style.

src/examples/design/icons/SvgSearch.tsx
```diff
@@ -29,85 +31,134 @@ const StyledSvgWrapper = styled.div<{ isAnswerBot?: boolean }>`
...
-      ));
-  }, [data, debouncedInputValue, StyledCol]);
+      .map(edge => {
+        return <Icon key={edge.node.name} {...edge} />;
+      });
+  }, [data.edges, inputValue]);
+
+  const onInputChange = debounce((event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+  }, 300);
```
hzhu, line 113: The debounced function is unnecessarily created each time a user types (state change). I think this was prevented previously by using `useCallback`. Another way to address this is to store the debounced function inside a `ref`. To ensure there aren't memory leaks, the component should clean up the debounced function:

```jsx
const Comp = () => {
  const [state, setState] = useState();
  const debouncedFn = useRef();

  useEffect(() => {
    debouncedFn.current = debounce(setState, 250);
    
    return () => {
      debouncedFn.current.cancel();
    }
  }, []);

  // ...
}

```

hzhu (changes requested):

src/examples/design/icons/SvgSearch.tsx
exelarios: Ah, interesting. Does using `useCallback` also resolve the issue with debounce having memory leaks?

src/pages/design/icons/library.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+Icon tokens represent specific concepts in Zendesk products.
+Before adding an icon to your project, check the list below to see if the concept has an icon tokenassociated with it.
+To suggest a new token, please take a look at our [icon token decision tree](https://miro.com/app/board/uXjVOcPRJo0)
+and fill out a Jira template if your request meets the criteria for addition. -->
+
+### All Icons
+
+This is a list of all the icons in the Garden library, independent of context.
+Use these icons only if you can’t find a suitable token.
```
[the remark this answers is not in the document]
exelarios: Haha, I should comment it out for now until the tokens are implemented.

src/examples/design/icons/SvgSearch.tsx
hzhu: Yeah `useCallback` and `useMemo` should also work because they cache the debounced function. A `useEffect` is still necessary to perform clean up via `debouncedFn.cancel()`.

> The debounced function comes with a cancel method to cancel delayed func invocations

src/components/MarkdownProvider/components/BestPractice.tsx
exelarios: I do agree, we should remove the provided padding on our end to allow designers to have more control of the examples. I'll remove the extra prop and leave `Don't use 32px icons for UI` as to how it was before.

I'm guessing this can be potentially a separate PR to address the padding removal from the examples?

For context:
One of the examples had a filled background for the image which looks a bit odd due to the padding.
image

With the `imageBackgrounfColor`
image

src/pages/design/icons/index.mdx
```diff
@@ -0,0 +1,116 @@
...
+#### Styles
+
+import IconStyles from '../../../examples/design/icons/IconStyles.tsx';
+import IconStylesCode from '!!raw-loader!../../../examples/design/icons/IconStyles.tsx';
+
+<CodeExample code={IconStylesCode}>
+  <IconStyles />
+</CodeExample>
+
+Icons also come in two styles: outline and filled.
```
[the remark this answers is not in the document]
diana-oum: @bethanybeech anything to weigh in here on Henry's good question? Looking at the old Garden icons page again, stroke and fill are what the icons are called so I think it's a good call out.

src/examples/design/icons/IconStyles.tsx
diana-oum: @hzhu Major +1. I like your suggestion

src/pages/design/icons/index.mdx
```diff
@@ -0,0 +1,116 @@
...
+    <p>Use 32px icons for illustration.</p>
+  </Do>
+</BestPractice>
+
+<BestPractice>
+  <Dont
+    imageBackgroundColor="#D2D5D8"
+    imageSource={props.data.uiIconSize.childFile.childImageSharp.gatsbyImageData}
+  >
+    <p>Don’t use 32px icons for UI.</p>
```
diana-oum, line 63: @bethanybeech Don't recall if I saw this content in the last iteration we worked on together but regardless, can we change this to something like this "Don't use 32px icons for functions in the user interface." I want to be more clear here since illustrations can technically be in the UI. What are your thoughts Bethany?

src/pages/design/icons/library.mdx
```diff
@@ -0,0 +1,76 @@
...
+To suggest a new token, please consider the following:
+
+- Is my icon repeatable? Can it be used in at least 3 places?
+  If not, consider using a local icon rather than adding to the Garden library.
+
+- Will it be used across mutliple products?
+
+- Is it unique? Is it unlike any exisiting Garden icon?
+
+To suggest an icon, please post in [#garden-seeds](https://app.slack.com/client/T024F4EL1/C02JQBLQX5W/)
```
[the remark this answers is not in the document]
diana-oum: Much of this can be housed internally in my experience but that's if the design system team has good relations with product teams and solid communication channels. Imo the website taking on this burden more since we don't have that culture as of now.

src/pages/design/icons/index.mdx
```diff
@@ -0,0 +1,116 @@
...
+<CodeExample code={IconShowcaseCode}>
+  <IconShowcase />
+</CodeExample>
+
+### Usage
+
+Garden’s icon library is implemented using SVG. Icons come two sizes: 12px and 16px.
+The 16px icons can also be doubled to create 32px icons.
+
+#### Styles
```
diana-oum, line 19: @bethanybeech What are your thoughts on making Styles an H3 and have it moved below 32px? The Usage description makes it read like we're going to just talk about sizes so let that come first and then Styles so the description flows "Icons also come in two styles..."

Also, thinking on this further, do you think 12px and 16px, 32px, and Styles should be H4s nesting under the H3 Usage?

diana-oum (changes requested):
I tagged Bethany in some of my comments so want her to weigh in before implementing those changes. Looking good so far!

src/examples/design/icons/IconShowcase.tsx
exelarios: I should be using `DEFAULT_THEME.palette.grey[600]` instead?

src/components/MarkdownProvider/components/BestPractice.tsx
hzhu: Oh I see, the padding is an issue so the image can't be re-sized to fit to the edges of the border. Seems like that's the intentional design of the component: image fits within the padding. Coloring the background gray to make the padding appear as part of the image is breaking that intention. Plus, the API for achieving that isn't great.

Personally I would a) abide by the current intention and leave the white spaced padding showing or b) I would adjust the image container to have zero padding and all images fit the container. Option b) is a separate task that needs team discussion.

src/examples/design/icons/IconShowcase.tsx
jzempel: @exelarios we'll tackle this together offline along with the grid re-work.

hzhu: No because that would still directly use the `PALETTE` object. Could use `const theme = useTheme();`. This isn't a request for change. I was more so curious if `theme` should be leveraged where possible in examples. Looks like it doesn't have to be: https://garden.zendesk.com/components/avatar#shape

Gonna resolve this comment 👍🏽 .

src/pages/design/icons/library.mdx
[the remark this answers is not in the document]
bethanybeech: Sorry, the language about this process is out of date definitely needs to be updated. That said, I think having some reference to our internal [Confluence page](https://zendesk.atlassian.net/wiki/spaces/GARDEN/pages/5379294927/Making+Garden+icons+revised+Q1+2022) is vital here. We can't let the open source nature of this project disadvantage our home team.

src/pages/design/icons/index.mdx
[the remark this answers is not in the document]
bethanybeech: Sure, let's keep it consistent.

src/pages/design/icons/library.mdx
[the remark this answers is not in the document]
bethanybeech: Whoops- I think there may have been a misunderstanding here. I think we should list the tokens on this page, even if they're not yet reflected in code. Designers and product owners can still use this page to reference which token goes with which icon, even if there is no code equivalent.

src/components/MarkdownProvider/components/BestPractice.tsx
bethanybeech: I asked him to do this. Is there a reason why changing the background color is bad? I think it looks better and I assume it isn't too disruptive to the structure of the page.

src/pages/design/icons/index.mdx
bethanybeech: Yep, I think something may have gotten mixed up here. In my comp, 12px and 16px, 32px, and Styles are all nested under the H3 Usage.

bethanybeech:
I commented on all open questions. Everything looks good! Two things I'd like to note:

1) I think we can include Icon Tokens as a section in the docs without actually making the actual tokens/classes in code. Having them listed out in a public page that's easily accessible will be super useful for designers, product owners, and pretty much anyone that need a quick reference for tokens. 

2) I just realized that the Icon Contribution language is out of date based on the revisions that Steph and I have been making to the contribution process. The current process is outlined [here](https://zendesk.atlassian.net/wiki/spaces/GARDEN/pages/5379294927/Making+Garden+icons+revised+Q1+2022). I am out this coming week, so if @diana-oum wants to/has time to update the language on this page to reflect those changes, I will happily defer to her expertise. Otherwise, we can work on revising this language when I return. Let's make sure we hold off on releasing this page until the language is updated- we wouldn't want to confuse anybody.

src/components/MarkdownProvider/components/BestPractice.tsx
hzhu: I agree I think it looks better. I think changing the background color to achieve it is bad because the `Dont` component was built to have padding around images. It looks like the reason for having padding was possibly flawed, so that should be looked into. Also, changing the background color could mislead the user when:

1) A user attempts to save an image. They right click the background color to save the image expecting the ability to right-click-and-save, it won't work since the clicked area is not part of the image.

2) A user saves the image. The saved image does not reflect what they saw on the website. The background color is much less in the image they saved.

I assume website visitors have no intention to save images, but these issue two points reveal flaws with the approach of changing the background color to achieve what the image should originally look like. The background color approach could be a temporary fix and there could be a separate task to investigate this issue.

src/examples/design/icons/SvgSearch.tsx
```diff
@@ -29,85 +31,148 @@ const StyledSvgWrapper = styled.div<{ isAnswerBot?: boolean }>`
...
+      </StyledIconWrapper>
+    </StyledCol>
+  );
+};
 
-  const StyledRow = styled(Row).attrs({ forwardedAs: 'ul' })``;
+export const SvgSearch: React.FC<ISvgSearchProps> = ({ data, searchEnabled }) => {
+  const [inputValue, setInputValue] = useState('');
+  const [collapsed, setCollapsed] = useState(true);
+  const debounceRef: any = useRef();
```
hzhu, line 89: Looks like this `ref` can be typed out to help the TS compiler:

```tsx
import type { DebouncedFunc } from 'lodash'; 

type ChangeHandler = (event: React.ChangeEvent<HTMLInputElement>) => void

const debounceRef = useRef<DebouncedFunc<ChangeHandler>>();

useEffect(() => {
  debounceRef.current = debounce((event) => {
    setInputValue(event.target.value);
  }, 300);

  return () => {
    if (debounceRef.current) {
      debounceRef.current.cancel();
    }
  };
}, []);


```

hzhu (approved):
LGTM, pending copy+design comments and grid re-work with @jzempel.

src/examples/design/icons/SvgSearch.tsx
exelarios: Oh nice catch, forgot to set a type before pushing the changes. 😅

smritimadan:
> 1. I think we can include Icon Tokens as a section in the docs without actually making the actual tokens/classes in code. Having them listed out in a public page that's easily accessible will be super useful for designers, product owners, and pretty much anyone that need a quick reference for tokens. I actually assumed that a list of tokens would be part of this release of this page, so apologies if that got lost in translation. If we can get these on the page for this release, I think it would be extremely useful, since we've now updated all of our internal process to include them, and people will expect to see them.

Hey @bethanybeech ! I had split the tokenization work into a separate ticket after getting your confirmation that we can publish the rest of icon restructuring without it, considering the unknowns around it. I'd like to stick to that approach if possible, so we can iterate and avoid blocking this PR! 

> 2. I just realized that the Icon Contribution language is out of date based on the revisions that Steph and I have been making to the contribution process. The current process is outlined [here](https://zendesk.atlassian.net/wiki/spaces/GARDEN/pages/5379294927/Making+Garden+icons+revised+Q1+2022). I am out this coming week, so if @diana-oum wants to/has time to update the language on this page to reflect those changes, I will happily defer to her expertise. Otherwise, we can work on revising this language when I return. Let's make sure we hold off on releasing this page until the language is updated- we wouldn't want to confuse anybody.

I have some questions around the timing of these revisions and how we could reduce this kind of rework at PR review stage in the future. We can talk about it offline!

Francois-Esquire (commented):
Thinking long term, it would be great to have parity with `zendesk-garden/svg-icons` project and extract documentation from a single source - however that may turn out (md fragments, etc).

src/components/MarkdownProvider/components/BestPractice.tsx
Francois-Esquire: Hmm I may be missing something here... where is `imageBackgroundColor` actually implemented?

src/examples/design/icons/SvgSearch.tsx
```diff
@@ -29,85 +31,148 @@ const StyledSvgWrapper = styled.div<{ isAnswerBot?: boolean }>`
...
               <XL>No icons found</XL>
             </StyledCol>
-          )}
-        </StyledRow>
+          </StyledRow>
+        )}
+        {collapsed && (
+          <StyledRow
+            css={css`
+              margin-bottom: ${p => p.theme.space.lg};
```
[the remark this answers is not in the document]
exelarios: Oh yeah. Good catch!

src/pages/design/icons/library.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+To suggest a new token, please consider the following:
+
+- Is my token repeatable? If the action will only appear once or very few times, using a one-off icon is fine.
+
+- Will it be used across multiple products?
+
+- Can it be consistently tied with a concept? Is there a clear action?
+
+If you can answer “yes” to these questions, please post in [#garden-seeds](https://app.slack.com/client/T024F4EL1/C02JQBLQX5W/) with the header “Token Request” and the Garden team will get back to you! -->
```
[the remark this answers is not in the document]
smritimadan: Wondering if we should just remove it (instead of commenting), just to keep it clean 😬?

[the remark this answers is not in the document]
exelarios: Actually, you're right. It would definitely make it more readable.

smritimadan:
> Thinking long term, it would be great to have parity with `zendesk-garden/svg-icons` project and extract documentation from a single source - however that may turn out (md fragments, etc).

Totally agree with this. Along similar lines, is there additional value to having the installation instructions repeated on the website? Could we not just link to the `zendesk-garden/svg-icons` project for the installation instructions?

src/pages/design/icons/library.mdx
```diff
@@ -0,0 +1,77 @@
...
+  If not, consider using a local icon rather than adding to the Garden library.
+
+- Will it be used across mutliple products?
+
+- Is it unique? Is it unlike any exisiting Garden icon?
+
+To suggest an icon, please post in [#garden-seeds](https://app.slack.com/client/T024F4EL1/C02JQBLQX5W/)
+with the header “Icon Request,” along with a description of your suggested icon and its associated token.
+
+<!-- markdownlint-enable line-length -->
```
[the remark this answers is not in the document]
exelarios: Oh yeah, that shouldn't be there. 😅
I'll have it removed.

src/examples/design/icons/SvgSearch.tsx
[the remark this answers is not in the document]
Francois-Esquire: Yes, not a blocker was just wondering if there were specific reasons to using the `css` prop.

src/pages/design/icons/index.mdx
```diff
@@ -0,0 +1,113 @@
...
+  </Dont>
+</BestPractice>
+
+### Best practices
+
+#### Use tooltips with interactive icons
+
+<BestPractice>
+  <Do imageSource={props.data.tooltipLabelIcon.childFile.childImageSharp.gatsbyImageData}>
+    <p>Use tooltips to label icons when they are used as actions. </p>
```
diana-oum, line 70: Double checking here: Is it just for actions? Are they used to simply identify things as well like concepts or navigation? 
image
To be inclusive of that, I'm thinking the section title should be "Include tooltips for interactive icons" and the image description to be "Use tooltips to clearly label interactive icons for accessibility." 
@bethanybeech or someone else who has been longer on the Garden team: that's the reason for these tooltips, correct? I feel like I heard Allison mention it's for accessibility.

diana-oum (changes requested):
Left a comment on a small content change.

Also, now that we don't have the contribution section on the Icon Library page, the Table of Contents on the right only has one things listed which looks strange. I don't think it needs to show up for that page since there's only one section now

src/pages/design/icons/index.mdx
bethanybeech: Sure, that makes sense to me.

exelarios:
> Left a comment on a small content change.
> 
> Also, now that we don't have the contribution section on the Icon Library page, the Table of Contents on the right only has one things listed which looks strange. I don't think it needs to show up for that page since there's only one section now

Yeah I agree it looks unnecessary to have a table of contents for only one listing. I will make a ticket to address this separately.

diana-oum (approved):

bethanybeech (approved):

Francois-Esquire (approved):

PR #352, opened 2022-02-18
feat(icons): add token metadata

Francois-Esquire:
## Description

This PR adds a hidden `<p />` tag to the icons with a token that it is linked to.

## Detail

The internal workings of this PR takes a CSV list of tokens and parses it to create a map. The map is then used in the `gatsby-transformer-garden-svg` plugin to add a field `token` to the icon file node and use that in the `SvgSearch` component to render an invisible `<p />` tag with the token. We can use this `<p />` tag to query `document.querySelectorAll('#main-content p')` with the expectation that Algolia docsearch will use our text selector `#main-content p` to extract and index the token in the `innerHTML` of each. 

## Checklist

### DIFF
--- a/plugins/gatsby-transformer-garden-svg/gatsby-node.js
+++ b/plugins/gatsby-transformer-garden-svg/gatsby-node.js
@@ -8,6 +8,8 @@
 const { createNodeHelpers } = require('gatsby-node-helpers');
 const { optimize } = require('svgo');
 
+const iconTokens = require('./icon-tokens');
+
 const config = {
   plugins: [
     {
@@ -34,7 +36,7 @@ const parseSvg = svgContent => {
 
 exports.onCreateNode = async ({
   node,
-  actions: { createNode, createParentChildLink },
+  actions: { createNode, createNodeField, createParentChildLink },
   loadNodeContent,
   createNodeId,
   createContentDigest
@@ -59,6 +61,11 @@ exports.onCreateNode = async ({
     ...parsedContent
   });
 
+  createNodeField({
+    node,
+    name: 'token',
+    value: iconTokens.get(node.name)
+  });
   createNode(svgNode);
   createParentChildLink({ parent: node, child: svgNode });
 };
--- a/plugins/gatsby-transformer-garden-svg/icon-tokens.csv
+++ b/plugins/gatsby-transformer-garden-svg/icon-tokens.csv
@@ -1,0 +1,45 @@
+Name,Icon
+Add New,plus
+Alert Error,alert-error
+Alert Info,info
+Alert Success,check-circle
+Alert Warning,alert-warning
+Attachment,paperclip
+Avatar,user-solo
+Back,chevron-left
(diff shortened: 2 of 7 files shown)


Francois-Esquire:
Moved PR out of draft for deployment.

jzempel (commented):
Was thought given to storing these tokens in proximity with the icons under https://github.com/zendeskgarden/svg-icons? Also, was a JSON format considered vs CSV? Finally, how does storing a capitalized icon name that is the same as the icon itself assist with tokenized search?

Before we go through with any of this, it seems like one temporary hard-coded hidden paragraph should be added to the icons page with a unique word (such as "alogolia") and deployed to garden.zendesk.com so that we can observe the search treatment. I have a hard time believing that they are going to allow a hidden paragraph to be indexed – due to how search results are displayed in the drop down.

Francois-Esquire:
> Was thought given to storing these tokens in proximity with the icons under https://github.com/zendeskgarden/svg-icons? Also, was a JSON format considered vs CSV? Finally, how does storing a capitalized icon name that is the same as the icon itself assist with tokenized search?

It's possible to store tokens in the `svg-icons` project however, how would we use this in the `website`? It doesn't feel correct to place this with the distribution IMO. JSON or CSV could work; the reason CSV was used for this example was decided by the original source of the tokens (with some modifications). Overall this is something we need to iron out as to how we structure and store icon tokens, as well as how we continue to feed updates as more tokens are added, etc.

For the capitalized icon name, I believe that is coincidentally the token given to that icon. Originally there were alternate search terms in the original source that were omitted for simplicity however we may need to clarify with @bethanybeech here.
 
> Before we go through with any of this, it seems like one temporary hard-coded hidden paragraph should be added to the icons page with a unique word (such as "alogolia") and deployed to garden.zendesk.com so that we can observe the search treatment. I have a hard time believing that they are going to allow a hidden paragraph to be indexed – due to how search results are displayed in the drop down.

I've added #362 to allow us to observe if the hidden element will be indexed or not.

jzempel:
> It's possible to store tokens in the `svg-icons` project however, how would we use this in the `website`?

I would think it could be imported JSON. Storing this metadata directly with the icons as a source-of-truth makes sense as 1) it can be reused in other contexts, 2) it is maintained directly with the creation/modification/deletion of the icons themselves.

Francois-Esquire:
> > It's possible to store tokens in the `svg-icons` project however, how would we use this in the `website`?
> 
> I would think it could be imported JSON. Storing this metadata directly with the icons as a source-of-truth makes sense as 1) it can be reused in other contexts, 2) it is maintained directly with the creation/modification/deletion of the icons themselves.

@jzempel I understand your logic behind this - Only thing is, tokens are only relevant with the website documentation now (I don't see any other use cases at the moment), could you explain more on which other contexts would use this?

The JSON structure makes complete sense. I will update this.

Francois-Esquire:
Closing in favor of #366 .

PR #366, opened 2022-03-14
feat(icons): token representation of icons

exelarios:
## Description
This PR adds tokens onto the Library section of icons. This PR also addresses the implementation of adding a hidden `<p/>` tag to the corresponding token to allow alternative terms to be queried by Alogolia (thanks @Francois-Esquire for looking into that).

## Detail
Tokens:
image

Hidden `<p/>` for alternative icon terms:
image

## Checklist


plugins/gatsby-transformer-garden-svg/tokens.json
```diff
@@ -0,0 +1,232 @@
+[
```
jzempel, line 1: Similar to my previous [comment](https://github.com/zendeskgarden/website/pull/352#pullrequestreview-890298772), I believe this information belongs co-located with and exported from `svg-icons`. This makes more sense from a maintenance and resusability POV. I would also like to take some time to go over the format together. It feels like these should be key-values, keyed by the icon prefix slug.

jzempel: However, before we take a run at the JSON placement/format, I'd like to get confirmation from @diana-oum on whether we're using Garden's sentence case for name [capitalization](https://garden.zendesk.com/content/grammar#capitalization).

diana-oum: In Bethany's mock, it's sentence case! @exelarios

exelarios: Ah, thanks @diana-oum! I didn't catch the capitalization.

@jzempel 
Hmm, I will consider moving the tokens into `svg-icons` since it does make more sense to have them there. I'll have a look into how I would move the JSON into the `svg-icons`.

The JSON is formatted as an array of Objects instead of key-values is because the SVG itself can point to multiple tokens; thus the icon SVG (e.g `plus-stroke`) name isn't unique. I'm not entirely sure what you mean by `keyed by the icon prefix slug`?

Do you mean something like this? (e.g `info-stroke`)
```
"info": {
 "name": "Alert Info",
 "alt": ["information", "I", "circle"]
}
```

jzempel: yes, your example is the key-value format I'm suggesting.

I also think that once this gets over to `svg-icons` it might make more sense to export as a JS `const`. Once you get that work started, please connect directly with me so we can discuss. 🙏

exelarios: I tried that at first but realized the SVG name points to multiple token names. That's the reason why I went with an array of Objects. Not sure, if I'm overlooking anything?
image

hzhu: I agree that the icons should be represented as an object and [normalized](https://redux.js.org/usage/structuring-reducers/normalizing-state-shape#designing-a-normalized-state). In this approach, an entity (icon) is keyed by its unique identifier (e.g. `info-stroke`). The full SVG icon file name should be used as the identifier since keys in a JavaScript object must be unique, and the object must account for both stroke and fill variants. Also, icon names (a new concept in Garden?) should be represented by an array since multiple names can belong to a unique icon.

I imagine the normalized object representation would be like:

```js
// In some central place, source of truth
const icons = {
  "chevron-left": {
    names: ["Hide Left", "Back"],
    tokens: []
  },
  "plus-stroke": {
    names: ["Add New"],
    tokens: []
  },
  "info-stroke": {
    names: ["Alert Info"],
    tokens: ["information", "i", "circle"]
  },
  ...restIcons
};
```

Given the normalized representation, UI data can then be derived from it:

```js
import icons from 'source-of-truth';

const iconDataForGatsby = Object.keys(icons).reduce((accumulator, identifier) => {
  const { names, tokens } = icons[identifier];
  return [
    ...accumulator,
    ...names.map((name) => ({
      icon: identifier,
      tokens,
      name
    }))
  ];
}, []);
```

plugins/gatsby-transformer-garden-svg/tokens.json
```diff
@@ -0,0 +1,232 @@
...
+  },
+  {
+    "name": "Attachment",
+    "icon": "paperclip-stroke",
+    "alternatives": ["file"]
+  },
+  {
+    "name": "Avatar",
+    "icon": "user-solo-stroke",
+    "alternatives": ["profile", "person", "user"]
```
hzhu, line 35: Is "profile", "person", and "user" supposed to be tokens? If so, would it make more sense to name this value "tokens" instead of "alternatives"?

Francois-Esquire: Good question, I'm not sure if they are alternate terms or tokens. @exelarios @bethanybeech any idea on this?

plugins/gatsby-transformer-garden-svg/tokens.json
```diff
@@ -0,0 +1,232 @@
...
+    "alternatives": []
+  },
+  {
+    "name": "Alert Error",
+    "icon": "alert-error-stroke",
+    "alternatives": ["circle", "exclamation point"]
+  },
+  {
+    "name": "Alert Info",
+    "icon": "info-stroke",
```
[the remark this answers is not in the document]
Francois-Esquire: If I'm not mistaken (and with some exceptions), `*-stroke` would be the "go-to" for mapping icon tokens.

plugins/gatsby-transformer-garden-svg/tokens.json
exelarios: JZ suggested the key name for alternate terms to be `synonyms`. Any thoughts on that too?

diana-oum: It's in the proposed garden icon tokens google sheet, those "alternatives" words come from the "metadata" column. "Profile", "person", "user" are words Bethany and I came up with to account for other search keywords that a user would think of to put in to search for the icon.

In my opinion, a more accurate single word name for them is "keywords" or more specifically "search-keywords"

hzhu: What is a "token"? And which line represents the token identifier?

I guess "Avatar" is the "token" here? If so then yeah "alternatives" or "synonyms" seem like a good key for `["profile", "person", "user"]`. But then shouldn't the key "name" be "token" instead since that _is_ the token?

plugins/gatsby-transformer-garden-svg/tokens.json
exelarios: Thanks @hzhu for thinking through this and sharing your approach. I love how you used the `reduce` function. 😆

Just synced up with @jzempel and he bought up a great point that there are currently three sources of truth; Figma, Google Sheets, and JSON if we end up with our current approach. So the new approach is directly reading tokens off from Figma where that can be the source-of-truth for the icon tokens.

Since there's a many-to-one relationship with the icon/slug, we're going with the naming of the tokens to the key identifier for the token.

JZ proposed the data be structured something like this:
```js
"add new": {
 "icon": "plus-stroke",
 "synonyms": []
}
```

plugins/gatsby-transformer-garden-svg/tokens.json
diana-oum: Big plus one to Henry. Having more literal and clear labels will help reduce the confusion like name --> token

plugins/gatsby-transformer-garden-svg/tokens.json
hzhu: I see, so the base schema is a token, not an icon. Essential each key (e.g. add new) is mapped to a token identifier. Yeah that structure makes sense to me.

plugins/gatsby-transformer-garden-svg/tokens.json
hzhu: Thanks for the clarification. Yeah `synonyms` or `keywords` seem good 👍🏽 .

exelarios: Thanks for clearing that up. I feel like we had different definitions of what `token` actually was.

plugins/gatsby-transformer-garden-svg/gatsby-node.js
```diff
@@ -5,9 +5,89 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
+const fetch = require('node-fetch');
 const { createNodeHelpers } = require('gatsby-node-helpers');
 const { optimize } = require('svgo');
```
jzempel, line 10: Are these imports used?

plugins/gatsby-transformer-garden-svg/gatsby-node.js
```diff
@@ -5,9 +5,89 @@
...
+
+  const file = await fetch(`${baseURL}/files/${fileId}/nodes?ids=${nodeId}`, { headers });
+  const { nodes } = await file.json();
+  const frame = nodes[Object.keys(nodes)[0]];
+  const { children } = frame.document;
+
+  for (const node of children) {
+    if (node.type === 'COMPONENT') {
+      // eslint-disable-next-line prefer-named-capture-group
+      const name = node.name.replace(/\b(\s+token)\b/u, '').toLowerCase();
```
jzempel, line 30: Prefer adding the non-capturing group `?:` notation rather than disabling the eslint rule

plugins/gatsby-transformer-garden-svg/gatsby-node.js
```diff
@@ -5,9 +5,89 @@
...
+      const id = node.id;
+      const iconId = node.children[0].componentId;
+      const icon = node.children[0].name
+        .split('-')[0]
+        .trim()
+        .replace('_', '')
+        .replace(/\s/gu, '-')
+        .toLowerCase();
```
jzempel, line 38: It would be good to preface this parsing block with a comment that defines the expected format of the data and the particular information that this is attempting to extract.

Have you tested this against malformatted data strings? How is this expected to gracefully fail?

jzempel (changes requested):
@exelarios there seems to be a fair bit of brute-force string `split` parsing with little/no graceful degradation (console warnings/errors) for unexpected data formats. Since we're parsing human-entered data here, the algorithms need to test for variable and ill-formatted data. Often this can be achieved by strong regex testing/matching/capturing groups.

plugins/gatsby-transformer-garden-svg/gatsby-node.js
exelarios: Yeah, those imports are being used.

`createNodeHelpers` is used to return the `createNodeFactory` and `optimize` is used within `parseSvg`.

plugins/gatsby-transformer-garden-svg/gatsby-node.js
exelarios: I'll definitely change it so it uses `?:` instead of `\b`.

gatsby-config.js
```diff
@@ -131,7 +131,14 @@ module.exports = {
     `gatsby-transformer-sharp`,
     `gatsby-plugin-image`,
     `gatsby-transformer-yaml`,
-    `gatsby-transformer-garden-svg`,
+    {
+      resolve: 'gatsby-transformer-garden-svg',
+      options: {
+        figmaApiToken: process.env.FIGMA_TOKEN,
+        fileId: '6g87L4FdKZTA3knt3Rsfdx',
```
jzempel, line 138: Do we know why this `fileId` does not match the one currently in use within `gatsby-config.js`? As I recall, all website assets should be coming from `'HifWgSrdeTlMMXUrcnrAAH'` ... is that right, @hzhu?

bethanybeech:
Hey @exelarios it looks like what's currently deployed might not be correct- I'm only seeing three icon tokens. I also think we can remove the "view all icons" button for the tokens section and just display all the tokens in full. There aren't a lot of items in that section, so it shouldn't take up too much space to show them all.

exelarios:
@bethanybeech Gotcha. I'll remove the `view all icons` button along with the gradient when the collapse isn't needed. 

As for the icon not fully populating, I'm currently looking into that. This seems to only happen during our production environment; the icons populate just fine during a test environment.

I also tested through a personal deployment via Netlify and all the tokens populate with the exception of `camera` since that's not uploaded onto the `svg-icons` repo.
https://icon-tokens.netlify.app/

gatsby-config.js
exelarios: I'm not sure if we're going to have to make it an exception because it would be awkward of having a copy of tokens along with their icons inside of `HifWgSrdeTlMMXUrcnrAAH` and would defeat the purpose of having a single source. 

Unless the designers have a way to sync the tokens and the icons inside of the Figma assets file with the current location of the icons.

plugins/gatsby-transformer-garden-svg/gatsby-node.js
exelarios: Mb, I didn't test this against malformatted data strings. I'm going to add additional checks to ensure during build time the whole application doesn't explode. I'm also going to see if I can minimize the amount of split and replace function through the use of regex.

PR #372, opened 2022-04-04
feat(actions): add deployment workflow

Francois-Esquire:
## Description

Adds GitHub action workflow to deploy the website on demand.

## Detail

This PR adds a new workflow that can be dispatched from the "Actions" tab. It will run `test`, `build` and finally a deployment script that re-uses the existing `deploy` script under `utils/` to run. Two parameters can be passed to the job, the `branch` and `figmaToken`. This workflow will need the `FIGMA_TOKEN` to run successfully, as well as the `NETLIFY_TOKEN` and `NETLIFY_SITE_ID`.

## Checklist


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,58 @@
+name: Deployment
+
+on:
+  # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch
+  workflow_dispatch:
+    branch:
+      description: 'The branch to deploy'
+      default: 'main'
```
jzempel, line 8: Why would we ever manually deploy a branch other than `main`?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      # https://github.com/actions/setup-node/issues/182
+      - name: Install Yarn
+        run: npm install --global yarn
+
+      - name: Install
+        run: yarn
+
+      - name: Test
+        run: yarn test
```
jzempel, line 43: Why is `yarn test` ever needed on a manual deployment? The `main` branch always tests with "success" status.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Deploy Production
+        if: ${{ github.ref == 'refs/heads/main' }}
+        # https://github.com/zendeskgarden/scripts/tree/main/src/netlify/deploy#command
+        run: yarn garden netlify-deploy public --production --message "https://github.com/$GITHUB_REPOSITORY/commit/$GITHUB_SHA"
+
+      - name: Deploy Draft
+        if: ${{ github.ref != 'refs/heads/main' }}
+        # https://github.com/zendeskgarden/scripts/tree/main/src/netlify/deploy#command
+        run: yarn garden netlify-deploy public --message "https://github.com/$GITHUB_REPOSITORY/commit/$GITHUB_SHA"
```
jzempel, line 58: Draft deploys using Garden's netlify token should always be connect with PR SHAs. See https://github.com/zendeskgarden/website/deployments for details. Garden developers are expected to use the `garden netlify-deploy` CLI and personal tokens for any one-off deployments (which should be very rare – draft deployments should be connected with PR discussions). This option confuses Garden's deployment-of-record between PRs and commits.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,58 @@
...
+  workflow_dispatch:
+    branch:
+      description: 'The branch to deploy'
+      default: 'main'
+      required: false
+      type: string
+
+env:
+  NODE_ENV: development
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jzempel, line 14: Is the GITHUB_TOKEN used?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,58 @@
...
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    environment: production
+    steps:
+      - uses: actions/checkout@v3
+        ref: ${{ github.event.inputs.branch }}
+        # https://github.com/actions/checkout/blob/main/src/input-helper.ts#L103
+        submodules: 'TRUE'
```
jzempel, line 27: What is this `uses` doing?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Setup
+        uses: actions/setup-node@v3
+        with:
+          node-version: '14'
```
jzempel, line 32: Prefer `.nvmrc` reuse: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

.github/workflows/deploy.yml
Francois-Esquire: This originally assumed that `main` branch is not the only target..

.github/workflows/deploy.yml
Francois-Esquire: At the moment, no we don't need it but we will if we want to trigger a GH deploy dispatch event

.github/workflows/deploy.yml
Francois-Esquire: It is pulling in the targeted branch.

.github/workflows/deploy.yml
Francois-Esquire: Is this the only case for draft deployments? What would be an alternative here if we wanted to sign the message in a similar matter? Maybe by branch instead of commit to distinguish between deployment types..

.github/workflows/deploy.yml
Francois-Esquire: It gives us the option pilot branches on a live URL before the PR process or for e2e testing in the future. This GH workflow can be extended to be configured to trigger by other events (like automated scheduled/periodic testing).

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Install
+        run: yarn
+
+      - name: Test
+        run: yarn test
+
+      - name: Build
+        env:
+          NODE_ENV: production
+        run: yarn build
```
jzempel, line 48: Please reference `.circleci/config.yml` – `yarn` already has the build handled.

.github/workflows/deploy.yml
jzempel: Ok, then I think we need to get to a place where this action matches the heuristic established by `deploy.js`. There are options for accomplishing this:
- enhance deploy.js to accept a `branch` parameter (see `react-components` [yarn new](https://github.com/zendeskgarden/react-components/blob/main/utils/scripts/new.js) or [yarn tag](https://github.com/zendeskgarden/react-components/blob/main/utils/scripts/tag.js) for the preferred `commander` setup)
- use the [`garden github-deploy`](https://github.com/zendeskgarden/scripts/tree/main/src/github/deploy#readme) CLI directly (this has the disadvantage of skipping the bandwidth check)

.github/workflows/deploy.yml
Francois-Esquire: I think I misunderstood in the start, update to build once.

.github/workflows/deploy.yml
Francois-Esquire: Agreed. Since we decided to go with `scripts/deploy.js` offline, we can be sure all of that is ran. And since the `@actions/checkout` is pointing us to the right branch before installation, we can be assured it will be available in the `deploy.js` script.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: actions/checkout@v3
+        ref: ${{ github.event.inputs.branch }}
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'yarn'
+
+      - name: build
+        run: yarn
```
jzempel, line 32: Needs `--frozen-lockfile` flag (compare with [circleci](https://app.circleci.com/pipelines/github/zendeskgarden/website/2320/workflows/789290af-76c4-4d37-bf51-b7a413ed978b/jobs/5034/parallel-runs/0/steps/0-106)).

jzempel (changes requested):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
+name: Deployment
+
+on:
+  # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch
```
hzhu, line 4: > Adds GitHub action workflow to deploy the website on demand.

I think JZ asked something similar in a thread that was resolved, but I wasn't able to figure out when this workflow should be triggered. @Francois-Esquire could you provide clarification on why and when I would use this?

hzhu, line 4: > Adds GitHub action workflow to deploy the website on demand.

I think JZ asked something similar in a thread that was resolved, but I wasn't able to figure out when this workflow should be triggered. @Francois-Esquire could you provide clarification on why and when I would use this?

mtomcal:
I like seeing us move more towards Github Actions 👍 I am interested in seeing the answer to value proposition mentioned in https://github.com/zendeskgarden/website/pull/372#discussion_r853362956

.github/workflows/deploy.yml
Francois-Esquire: Ahh good call, will update.

.github/workflows/deploy.yml
Francois-Esquire: @hzhu sure, I think the most common scenario is when we have updates pending outside of the codebase (Figma for example) and we want to republish the website based on the new data. Does that give a better idea?

hzhu: Oh I see. This is an alternative to clicking "rerun CI workflow" for the `main` branch in CircleCI? Thanks for the clarification. Feel free to resolve this thread.

jzempel (approved):
I've added the actions secrets

PR #378, opened 2022-04-19
feat(chrome): Adds back 26px icons onto the website with textual context

exelarios:
## Description
Adds the 26px icons back into the website with a description of their usage.
Also, added a new prop for `svgSearch`, in order to control the collapsible. 

## Detail
This PR also adds a section to the Chome component page that redirects the user to the 26px icons.
image

The hyperlink will direct the user a library of chrome icons (26px).
image

## Checklist

### DIFF
--- a/src/examples/design/icons/SvgSearch.tsx
+++ b/src/examples/design/icons/SvgSearch.tsx
@@ -10,7 +10,7 @@ import { Grid, Row, Col } from '@zendeskgarden/react-grid';
 import debounce from 'lodash/debounce';
 import type { DebouncedFunc } from 'lodash';
 import { rgba } from 'polished';
-import { Field, MediaInput, Label } from '@zendeskgarden/react-forms';
+import { Field, MediaInput } from '@zendeskgarden/react-forms';
 import { ReactComponent as SearchStroke } from '@zendeskgarden/svg-icons/src/16/search-stroke.svg';
 import { Code, XL } from '@zendeskgarden/react-typography';
 import { Button } from '@zendeskgarden/react-buttons';
@@ -68,7 +68,8 @@ interface ISvgSearchProps {
   data: {
     edges: ISvgNodeProps[];
   };
-  collapsible: boolean;
+  inputPlaceholder?: string;
+  defaultCollapsed: boolean;
 }
 
 type ChangeHandler = (event: React.ChangeEvent<HTMLInputElement>) => void;
@@ -90,9 +91,14 @@ const Icon = (edge: ISvgNodeProps) => {
   );
 };
 
-export const SvgSearch: React.FC<ISvgSearchProps> = ({ data, collapsible, searchEnabled }) => {
+export const SvgSearch: React.FC<ISvgSearchProps> = ({
+  data,
+  defaultCollapsed,
+  inputPlaceholder,
+  searchEnabled
+}) => {
   const [inputValue, setInputValue] = useState('');
-  const [collapsed, setCollapsed] = useState(collapsible);
+  const [collapsed, setCollapsed] = useState(defaultCollapsed);
   const debounceRef = useRef<DebouncedFunc<ChangeHandler>>();
 
   const icons = useMemo(() => {
@@ -143,8 +149,11 @@ export const SvgSearch: React.FC<ISvgSearchProps> = ({ data, collapsible, search
                 margin-bottom: ${p => p.theme.space.lg};
               `}
             >
-              <Label>Search icons</Label>
-              <MediaInput start={<SearchStroke />} onChange={onInputChange} />
+              <MediaInput
+                start={<SearchStroke />}
+                placeholder={inputPlaceholder}
+                onChange={onInputChange}
+              />
             </Field>
           </Col>
         </Row>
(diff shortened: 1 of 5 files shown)


src/examples/design/icons/SvgSearch.tsx
```diff
@@ -68,6 +68,7 @@ interface ISvgSearchProps {
   data: {
     edges: ISvgNodeProps[];
   };
+  collapsible: boolean;
```
hzhu, line 71: This prop configures the default uncontrolled state, so a more suitable name is `defaultCollapsed`. This follows the general conventions of HTML & React elements:

```jsx
const Example = (
  <>
    <input type="text" defaultValue="hello world" /> 
    <input type="checkbox" defaultChecked /> 
  </>
)
```

hzhu (commented):
Is this ready for review? It looks like the PR checklist is not ready?

Also I don't think "readds" is a word. Maybe this could be updated to be "Adds back..." or "Adds...".

src/examples/design/icons/SvgSearch.tsx
exelarios: Gotcha, I'll make the changes. 😁

src/nav/design.yml
```diff
@@ -12,3 +12,5 @@
           title: Overview
         - id: /design/icons/library
           title: Library
+        - id: /design/icons/chrome-icons
+          title: Chrome Icons
```
diana-oum, line 16: Lowercase "i" in 'icons please! We use sentence case on the Garden website

src/pages/design/icons/chrome-icons.mdx
```diff
@@ -0,0 +1,37 @@
+---
+title: Chrome Icons
```
diana-oum, line 2: Same thing here, sentence case. Lowercase i for icons

Also, tangent: Icon overview and and icon library page titles aren't in sentence case either. Mind if you could catch those as well, Deric? Thank you!!

diana-oum (changes requested):

src/pages/design/icons/chrome-icons.mdx
exelarios: Sounds good! Thanks for the review!

Francois-Esquire (approved):
Pending everyones comments, I think it looks good otherwise!

diana-oum (approved):

src/examples/design/icons/SvgSearch.tsx
```diff
@@ -142,8 +149,11 @@ export const SvgSearch: React.FC<ISvgSearchProps> = ({ data, searchEnabled }) =>
                 margin-bottom: ${p => p.theme.space.lg};
               `}
             >
-              <Label>Search icons</Label>
```
hzhu, line 145: Any particular reason why this was removed?

Inputs are usually accompanied by labels so that screen readers can identify the form control to assistive devices. The input has an `aria-labelledby` that points to an `id` that doesn't exist in the DOM.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input
https://webaim.org/techniques/forms/controls#input

src/pages/design/icons/library.mdx
```diff
@@ -10,7 +10,7 @@ This is a list of all the icons in the Garden library, independent of context.
 
 import { SvgSearch } from '../../../examples/design/icons/SvgSearch';
 
-<SvgSearch data={props.data.svgIcons} />
+<SvgSearch data={props.data.svgIcons} inputPlaceholder="Filter all icons" />
```
diana-oum, line 13: This is a search and not a filter correct? I think the content should be "Search icons"

diana-oum (changes requested):

bethanybeech (approved):
Looks good! :)

diana-oum (approved):
:shipit: 

Bethany and I made alternate search terms for the other icons. We can do the same for these but wondering if engineering will have the bandwidth to implement before Design does the work cc: @smritimadan @nyckie

mtomcal:
I would take a look at https://github.com/zendeskgarden/website/pull/378#discussion_r856477712 but looks good to me other than that 👍

mtomcal (approved):

exelarios:
> I would take a look at https://github.com/zendeskgarden/website/pull/378#discussion_r856477712 but looks good to me other than that 👍

@mtomcal Oh shoot, thanks Michael for letting me know. For some reason, I didn't see Henry post.

src/examples/design/icons/SvgSearch.tsx
exelarios: Oh, interesting. I didn't know the `Label` accompanies the screen readers. I'll reach out to Bethany about why the `Label` isn't present `MediaInput`.

hzhu: Some inputs don't have visible labels (e.g. https://google.com). But the DOM should still include a label for screen readers. Also, the accessibility tree should make sense (e.g. `aria-labelledby` should not point to a non-existant id).

exelarios: Ohh gotcha. There's an option to have the `Label` hidden via props.
Thanks for catching that Henry and hinting me to the solution. 😆

hzhu (approved):

PR #391, opened 2022-05-11
fix(tables):  Provided a valid label for form fields

exelarios:
## Description
The checkboxes within the table now have a unique label to discern the specific purpose of each field.

## Detail
`Before`
Screen reader wasn't able to distinguish each field due to identical labels.
example of identical labels.

`After`
example of unique labels

## Checklist

### DIFF
--- a/src/examples/table/TableSelectRows.tsx
+++ b/src/examples/table/TableSelectRows.tsx
@@ -120,7 +120,7 @@ const Example = () => {
                       setFocusedRowIndex(index);
                     }}
                   >
-                    <Label hidden>Select ticket #{index}</Label>
+                    <Label hidden>Select ticket for {data[index].fruit}</Label>
                   </Checkbox>
                 </Field>
               </Cell>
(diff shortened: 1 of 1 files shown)


src/examples/table/TableSelectRows.tsx
```diff
@@ -120,7 +120,7 @@ const Example = () => {
                       setFocusedRowIndex(index);
                     }}
                   >
-                    <Label hidden>Select ticket</Label>
+                    <Label hidden>Select ticket #{index}</Label>
```
smritimadan, line 123: I think we should be even more descriptive still since "#1" would still not provide enough info to an AT user about the row being referenced. How about:

```suggestion
                    <Label hidden>Select ticket for Fruit #{index}</Label>
```

exelarios: Sounds good. Changes have been made.

smritimadan (approved):

Francois-Esquire (approved):

PR #452, opened 2022-12-01
fix(components/tables): refactor scroll example with sticky head

Francois-Esquire:
## Description

Uses `position: sticky;` with the `Head` from `@zendeskgarden/react-table` to accessibly bind an HTML table for the "Scroll" example.

## Detail

Screen Shot 2022-12-01 at 11 48 10 AM

Notable changes:
- The scrollbar has shifted to include the `Head` area; prior the scroll region did not include this area.

Opportunities to enhance the experience:
- use of `Caption` to give the `Table` a title
- Adding an `aria-label` to each `Row` to announce interactions (selectable)

## Checklist


geotrev:
It's probably fine here, but SRs have a hard time dealing with alternative displays or positions and might not announce or properly focus table rows/cells. :( might be worth testing how NVDA/VO interprets the table using quick nav/browse mode (luckily JAWS is pretty fault tolerant in comparison to the others).

src/examples/table/TableScroll.tsx
```diff
@@ -7,14 +7,18 @@
...
+  top: 0;
+  box-shadow: ${props => {
+    const boxShadow = props.theme.shadows.sm(getColor('neutralHue', 200, props.theme) as string);
 
-const StyledSpacerCell = styled(HeaderCell)`
-  padding: 0;
-  width: ${SCROLLBAR_SIZE}px;
+    return `${boxShadow.replace('2px', '1px')}`;
+  }};
+  background-color: ${props => props.theme.colors.background};
```
jzempel, line 21: The `box-shadow` here should be expressed as a themed [border](https://6388e41c6b71d01c2d65dcb5--zendeskgarden.netlify.app/components/theme-object#borders), since that is what it is acting as. Also, there is a problem with the thickened border under the header.

src/examples/table/TableScroll.tsx
```diff
@@ -32,30 +36,25 @@ const rowData: IRow[] = Array.from(Array(100)).map((row, index) => ({
 }));
 
 const Example = () => (
-  <div style={{ overflowX: 'auto' }}>
-    <Table style={{ minWidth: 500 }}>
-      <Head>
+  <div style={{ maxHeight: 500, overflowY: 'auto' }}>
+    <Table>
```
jzempel, line 40: Why was the `minWidth` removed?

jzempel (changes requested):
Beyond the styling [nit], I think there needs to be a larger discussion about rolling this back as a feature within `react-components` (as `<Head isSticky>` perhaps?). It's not ideal to be forcing this amount of customization on to the consumer.

jzempel:
> ...but SRs have a hard time dealing with alternative displays or positions and might not announce or properly focus table rows/cells. :( might be worth testing how NVDA/VO interprets the table using quick nav/browse mode

+1 re: this concern. Another reason to consider this work under https://zendeskgarden.github.io/react-components/?path=/story/packages-tables-table--table where we can push on the prop combinations.

src/examples/table/TableScroll.tsx
Francois-Esquire: No particular reason than when an extra `div` was removed.

src/examples/table/TableScroll.tsx
Francois-Esquire: I removed the border under the `Header` to correct for the thickened border. I also updated the `box-shadow` to follow the existing example from the `Table` pattern.

src/examples/table/TableScroll.tsx
```diff
@@ -7,14 +7,19 @@
 
 import React from 'react';
 import styled from 'styled-components';
-import getScrollbarSize from 'dom-helpers/scrollbarSize';
+import { getColor } from '@zendeskgarden/react-theming';
 import { Body, Cell, Head, HeaderCell, HeaderRow, Row, Table } from '@zendeskgarden/react-tables';
 
-const SCROLLBAR_SIZE = getScrollbarSize();
+const StyledHeaderRow = styled(HeaderRow)`
+  border-bottom: none;
```
jzempel, line 14:
```suggestion
  border-bottom-color: transparent;
```

It's very subtle, but there is undesired compaction that happens when a border is removed on a table with `border-collapse: collapse`.

src/examples/table/TableScroll.tsx
```diff
@@ -7,14 +7,19 @@
...
+`;
 
-const StyledSpacerCell = styled(HeaderCell)`
-  padding: 0;
-  width: ${SCROLLBAR_SIZE}px;
+const StyledHead = styled(Head)`
+  box-sizing: border-box;
+  position: sticky;
+  top: 0;
+  box-shadow: inset 0 -1px 0 ${props => getColor('neutralHue', 300, props.theme)};
```
jzempel, line 21:
```suggestion
  box-shadow: inset 0 -${props => props.theme.borderWidths.sm} 0 ${props => getColor('neutralHue', 300, props.theme)};
```

per https://github.com/zendeskgarden/website/pull/452#discussion_r1038360871

src/examples/table/TableScroll.tsx
```diff
@@ -7,14 +7,19 @@
...
-const SCROLLBAR_SIZE = getScrollbarSize();
+const StyledHeaderRow = styled(HeaderRow)`
+  border-bottom: none;
+`;
 
-const StyledSpacerCell = styled(HeaderCell)`
-  padding: 0;
-  width: ${SCROLLBAR_SIZE}px;
+const StyledHead = styled(Head)`
+  box-sizing: border-box;
```
jzempel, line 18:
```suggestion
```

I don't think this line is relevant.

src/examples/table/TableScroll.tsx
```diff
@@ -32,30 +33,25 @@ const rowData: IRow[] = Array.from(Array(100)).map((row, index) => ({
 }));
 
 const Example = () => (
-  <div style={{ overflowX: 'auto' }}>
+  <div style={{ maxHeight: 500, overflowY: 'auto' }}>
```
Francois-Esquire, line 36: Do we want to make this `overflow: auto` in both axis?
```suggestion
  <div style={{ maxHeight: 500, overflow: 'auto' }}>
```

jzempel (changes requested):

PR #464, opened 2023-01-31
fix(examples): return focus to trigger when drawer modal is closed

Francois-Esquire:
## Description

Adds the `restoreFocus` and `focusOnMount` props to the default `DrawerModal` example to return focus to the trigger element.

## Checklist

### DIFF
--- a/src/examples/drawer/DrawerDefault.tsx
+++ b/src/examples/drawer/DrawerDefault.tsx
@@ -20,7 +20,7 @@ const Example = () => {
     <Row>
       <Col textAlign="center">
         <Button onClick={open}>Open Drawer</Button>
-        <DrawerModal isOpen={isOpen} onClose={close}>
+        <DrawerModal isOpen={isOpen} onClose={close} restoreFocus focusOnMount>
           <DrawerModal.Header tag="h2">Tending a garden</DrawerModal.Header>
           <DrawerModal.Body>
             <Paragraph>
--- a/src/examples/drawer/DrawerDefault.tsx
+++ b/src/examples/drawer/DrawerDefault.tsx
@@ -20,7 +20,7 @@ const Example = () => {
     <Row>
       <Col textAlign="center">
         <Button onClick={open}>Open Drawer</Button>
-        <DrawerModal isOpen={isOpen} onClose={close} restoreFocus focusOnMount>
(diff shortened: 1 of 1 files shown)


src/examples/drawer/DrawerDefault.tsx
```diff
@@ -20,7 +20,7 @@ const Example = () => {
     <Row>
       <Col textAlign="center">
         <Button onClick={open}>Open Drawer</Button>
-        <DrawerModal isOpen={isOpen} onClose={close}>
+        <DrawerModal isOpen={isOpen} onClose={close} restoreFocus focusOnMount>
```
jzempel, line 23: This isn't working for me. When I pull the example into CodeSandbox, only `restoreFocus` is needed to restore keyboard focus back to the trigger button (just like the `Modal` example).

jzempel (changes requested):

src/examples/drawer/DrawerDefault.tsx
Francois-Esquire: Thats correct, I am in the process of investigating as to why this isn't working as expected and it looks like focus is reverting to `#gatsby-focus-wrapper` for some odd reason. I will continue to look more into this.

jzempel: Did you try in production mode? I seem to remember we might've seen some odd Gatsby stuff with this in development.

Francois-Esquire: It didn't seem to work in development mode as well...

jzempel (approved):

geotrev (approved):
LGTM

jzempel (changes requested):
On second thought, I think there's more to do here with unmounting + animation. When I matched the default `<Modal>` example, the `restoreFocus` works as intended.

Francois-Esquire:
> On second thought, I think there's more to do here with unmounting + animation. When I matched the default `<Modal>` example, the `restoreFocus` works as intended.

Yes, mounting/unmounting the component does work however the animation is completely void when doing so.

geotrev:
Latest deployment from my PR includes the drawer change: https://63ea6942dbe9e40052c2e721--zendeskgarden.netlify.app/

Focus is working correctly now. 👍🏻

PR #465, opened 2023-02-07
fix(forms): updates FileUpload & Tag examples with new prop patterns

geotrev:
## Description

Updates `File` and `Tag` (including `File.Close`, `File.Delete`, and `Tag.Delete`) component examples with latest accessible practices.

There are no copy updates.

## Detail

Main changes to `File, `Tag`, and `File.Close`, `File.Delete`, and `Tag.Close` include:
1. `tabIndex={-1}` is no longer applied (that happens automatically in close/delete sub-components)
2. `onKeyDown` handler added (to handle enter and space keys)
3. More descriptive `aria-label`s used (on `File`, `Tag`, and close/delete sub-components)
4. Tooltips wrap close/delete sub-components

Example before:

```js
<File type="pdf" aria-label="PDF file" tabIndex={0} onKeyDown={handleKeyDown}>
  Basics of gardening.pdf
  <Tooltip content="Stop upload">
    <File.Close aria-label="close" onClick={handleClick} tabIndex={-1} />
  </Tooltip>
  <Progress value={48} aria-label="Uploading Basics of gardening.pdf" />
</File>

<Tag tabIndex={0} onKeyDown={handleKeyDown}>
  <span>Algae</span>
  <Tag.Close onClick={() => alert('Tag dismissed via mouse')} />
</Tag>
```

Example after:

```js
<File
  type="pdf"
  aria-label="PDF file, press delete to stop upload"
  tabIndex={0}
  onKeyDown={handleFileKeyDown}
>
  Basics of gardening.pdf
  <Tooltip content="Stop upload">
    <File.Close
      aria-label="Stop upload"
      onClick={handleClick}
      onKeyDown={handleCloseKeyDown}
    />
  </Tooltip>
  <Progress value={48} aria-label="Uploading Basics of gardening.pdf" />
</File>

<Tag tabIndex={0} aria-label="Algae, press delete to remove" onKeyDown={handleTagKeyDown}>
  <span>Algae</span>
  <Tooltip content="Remove tag">
    <Tag.Close
      aria-label="Remove tag"
      onClick={() => alert('Tag dismissed via mouse')}
      onKeyDown={handleCloseKeyDown}
    />
  </Tooltip>
</Tag>
```

Note: the extra aria-labels are needed because of `useText`, to nudge folks on translation.

## Checklist


Francois-Esquire:
What was the reason to remove the `Tooltip` component from the examples?

geotrev:
> What was the reason to remove the `Tooltip` component from the examples?

The tooltip's trigger (close button) isn't navigable outside of specific screen reader usage (browse/quick nav modes). And in that scenario, SRs will provide a caption anyway.

src/examples/forms/file-upload/FileList.tsx
```diff
@@ -10,54 +10,75 @@ import { FileList, File } from '@zendeskgarden/react-forms';
...
   <Row justifyContent="center">
     <Col sm={5}>
       <FileList>
         <FileList.Item>
-          <File type="generic" aria-label="Generic file" tabIndex={0} onKeyDown={handleKeyDown}>
+          <File type="generic" aria-label="Generic file" tabIndex={0} onKeyDown={handleFileKeyDown}>
             Garden file
-            <Tooltip content="Stop upload">
-              <File.Close aria-label="close" onClick={handleClick} tabIndex={-1} />
-            </Tooltip>
```
jzempel, line 33: I think we need to retain this `Tooltip`. It is important for mouse accessibility – aka, levels the experience for both mouse & keyboard users.

src/examples/forms/file-upload/FileList.tsx
```diff
@@ -10,54 +10,75 @@ import { FileList, File } from '@zendeskgarden/react-forms';
...
+            <File.Close
+              aria-label="Press delete to remove"
```
jzempel, line 40: Please retain the original tooltip intent throughout. In this case the label should be "Stop upload" as that is the intended design for what we expect users to do with the `File.Close` interaction.

src/examples/forms/file-upload/FileList.tsx
```diff
@@ -80,21 +103,21 @@ const Example = () => (
...
           </File>
         </FileList.Item>
         <FileList.Item>
-          <File type="zip" aria-label="ZIP file" tabIndex={0} onKeyDown={handleKeyDown}>
+          <File type="zip" aria-label="ZIP file" tabIndex={0} onKeyDown={handleFileKeyDown}>
             Landscape.zip
-            <Tooltip content="Remove file">
-              <File.Delete aria-label="delete" onClick={handleClick} tabIndex={-1} />
-            </Tooltip>
+            <File.Delete aria-label="delete" onClick={handleClick} onKeyDown={handleCloseKeyDown} />
```
jzempel, line 120: Similar to previous comment, please use a "Remove file" label here

src/examples/forms/file-upload/FileUpload.tsx
```diff
@@ -49,30 +48,45 @@ const FileItem: React.FC<{ name: string; onRemove: () => void }> = memo(({ name,
...
-        </Tooltip>
+        {progress === 100 ? (
+          <File.Delete
+            aria-label="Press delete to remove"
+            onClick={onRemove}
+            onKeyDown={handleCloseKeyDown}
+          />
+        ) : (
+          <File.Close
+            aria-label="Press delete to remove"
```
jzempel, line 85: This should not be "delete" verbiage.

jzempel (changes requested):
@geotrev let me know if we should sync up on the `Tooltip` issue and the impact on these examples.

src/examples/forms/file-upload/FileList.tsx
geotrev: Ah, good point. I went to change one thing and accidentally chiseled it too much. :)

geotrev:
Worth calling out is File/Tag won't work properly until Thursday's release; Tooltip adds an additional tabs top to `File.Close`/`File.Delete`/`Tag.Close`. Doing a local symlink will show the true behavior until then.

src/examples/tags/TagsClose.tsx
```diff
@@ -9,20 +9,36 @@ import React from 'react';
...
+};
+
 const Example = () => (
   <Row>
     <Col textAlign="center">
-      <Tag tabIndex={0} onKeyDown={handleKeyDown}>
+      <Tag tabIndex={0} aria-label="Algae, press delete to remove" onKeyDown={handleTagKeyDown}>
         <span>Algae</span>
-        <Tag.Close onClick={() => alert('Tag dismissed via mouse')} />
+        <Tooltip content="Remove tag">
```
jzempel, line 35: Yikes. We _definitely_ cannot have a double nav on either `<Tag>` or `<File>`. This would add serious issues to product. I wonder if that brings us full circle to why we had these components structured as they originally were? 😬.

jzempel (changes requested):

src/examples/tags/TagsClose.tsx
geotrev: Perhaps my previous comment was super clear - there's only a double tab on the website, in this PR specifically, because of the changes in the PR itself (adding Tooltip to File.Close, File.Delete, and Tag.Close without the latest changes). That double tab will go away once we uptake the next react-components version on the website and I rebase this branch. Currently there isn't a double tab from File or Tag components.

jzempel: That's a relief. I'll re-review once `react-components` is released and updated here.

geotrev: Updated. I can split the dep updates to a separate PR if desired.

jzempel (approved):

PR #466, opened 2023-02-13
fix(examples): adds hidden prop to empty cell on draggable table rows example

geotrev:
## Description

✅ `v8.63.1` on `main`, branch rebased

Applies new `hidden` prop to empty cell header on Tables' draggable row example.

## Checklist

### DIFF
--- a/src/examples/table/TableDraggable.tsx
+++ b/src/examples/table/TableDraggable.tsx
@@ -188,7 +188,7 @@ const Example = () => {
       <Table>
         <Head>
           <HeaderRow>
-            <HeaderCell isMinimum />
+            <HeaderCell isMinimum hidden />
             <HeaderCell>Fruit</HeaderCell>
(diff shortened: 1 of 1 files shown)


src/examples/table/TableDraggable.tsx
```diff
@@ -188,7 +188,7 @@ const Example = () => {
       <Table>
         <Head>
           <HeaderRow>
-            <HeaderCell isMinimum />
+            <HeaderCell isMinimum hidden />
```
smritimadan, line 191: To hide this accessibly, won't we need some text for the Cell here?

jzempel: Yes, I'd recommend using "Draggable grip" as the content.

geotrev: Good call.

jzempel (approved):

Francois-Esquire (approved):

PR #480, opened 2023-04-03
feat(markdown/components): add img alt text prop for best practices e…

Francois-Esquire:
## Description

Add `imageAltText` prop for `BestPractice` components.

## Detail

Adding support for `alt` attribute in a `BestPractices` section for `Do` and `Dont` components. The prop that is exposed is `imageAltText` which is passed as the `alt` attribute for `img` or `<GatsbyImage>` component. The default value is an empty string if no value is provided as before.

## Checklist


jzempel (commented):
The do/don't sections were `alt=''` by design, since best practices should always provide supporting paragraph text which is accessible to screenreaders. I don't think we want to encourage a double-readout here with the addition of an `alt` tag.

Francois-Esquire:
@jzempel thanks for clarifying, closing!

PR #493, opened 2023-05-19
fix: re-adds pattern link to dnd pattern docs

geotrev:
## Description

Adds link to drag and drop pattern for dnd docs.

## Checklist

- [ ] ~~:wheelchair: analyzed via [axe](https://www.deque.com/axe/) and evaluated using VoiceOver~~
- [ ] ~~:memo: tested in Chrome, Firefox, Safari, Edge, and IE11~~

### DIFF
--- a/src/pages/patterns/drag-and-drop.mdx
+++ b/src/pages/patterns/drag-and-drop.mdx
@@ -8,7 +8,7 @@ description: >
 
 ## Anatomy
 
-[Drag and drop](/components/draggable)
+[Drag and drop](https://zendeskgarden.github.io/react-components/?path=/story/packages-drag-drop-patterns--drag-and-drop)
 interaction is common in workflows where the user wants to change the order or of items.
 It is also often used to build something by moving items around the screen,
 for example, a layout, report, or dashboard.


lucijanblagonic (approved):
Looks good! Thanks!

jzempel (changes requested):
Hmm, historically we prefer linking internally within the website rather than linking out. Was this decision reversed?

lucijanblagonic:
> Hmm, historically we prefer linking internally within the website rather than linking out. Was this decision reversed?

We did link to the [CKEditor](https://garden.zendesk.com/patterns/rich-text-editor#related-components) for the same reason, as this is where you test the pattern. I think the same reasoning applies here. Wouldn't do it for component pages, but in this case it makes sense to me, as this is talking about a pattern specifically, and the storybook example is the pattern example.

jzempel:
> ...the storybook example is the pattern example

In this particular case, we should link to https://garden.zendesk.com/components/draggable#default-2 instead, where consumers have access to extra goodness provided by the website (view/copy code, codesandbox, etc). In fact, the goal should probably be to rip the https://zendeskgarden.github.io/react-components/?path=/story/packages-drag-drop-patterns--drag-and-drop pattern now that the canonical example exists on garden.zendesk.com.

lucijanblagonic:
> @jzempel  In this particular case, we should link to https://garden.zendesk.com/components/draggable#default-2 instead, where consumers have access to extra goodness provided by the website (view/copy code, codesandbox, etc). In fact, the goal should probably be to rip the https://zendeskgarden.github.io/react-components/?path=/story/packages-drag-drop-patterns--drag-and-drop pattern now that the canonical example exists on garden.zendesk.com.

Your reasoning makes sense to me, but it seems a shame to lose those additional examples like Target indicators, which are not available in the component page. Any idea how to approach that?

Screenshot 2023-06-02 at 12 06 06

jzempel:
> ...it seems a shame to lose those additional examples like Target indicators, which are not available in the component page. Any idea how to approach that?

Ah, interesting. Does this suggest we're missing website demo(s)? If it doesn't fit neatly within a website demo, then I can see the value of linking out here.

lucijanblagonic:
> Ah, interesting. Does this suggest we're missing website demo(s)? If it doesn't fit neatly within a website demo, then I can see the value of linking out here.

We have the [Target indicator](https://garden.zendesk.com/components/draggable#indicator) static example, which I think works well to show how it looks, but it's not interactive. Linking to the storybook pattern is giving everyone a chance to discover/tweak the example & props. I think it's less about copying code. 

If it makes sense to create/embed an interactive demo into the pattern page, I am totally fine with that, but it's also going to be a tradeoff as pattern consumers won't have a quick way to "modify" the pattern to test it (without providing multiple interactive examples).  Maybe we need to "revisit" how we think about examples on pattern pages 🤔  but for now I think the easiest way is to just link to the pattern page, as proposed above.

jzempel:
> ...for now I think the easiest way is to just link to the pattern page, as proposed above

@lucijanblagonic got it - thanks for the clarification

jzempel (approved):

PR #496, opened 2023-05-26
feat(styles): update to new focus styles

Francois-Esquire:
## Description

Upgrading website components and `react-components` to the new focus style standards.

## Detail

- There is mixed usage of `Link` from `gatsby` and our `Anchor` component which causes an inconsistency of focus styles.

## Checklist

- [x] The PR will be moved from draft when `react-components` PR [#1548](https://github.com/zendeskgarden/react-components/pull/1548) is complete.


src/components/MarkdownProvider/components/Anchor.ts
```diff
@@ -5,37 +5,30 @@
...
+    margin-right: ${p => p.theme.space.base}px;
+    margin-left: -${p => math(`${p.theme.space.xs} * 2 + ${p.theme.iconSizes.md} + ${p.theme.space.base}`)};
     border-radius: ${p => p.theme.borderRadii.md};
     padding: 0 ${p => p.theme.iconSizes.md};
     color: transparent;
 
     /* stylelint-disable selector-max-specificity */
     ${p => mediaQuery('down', 'md', p.theme)} {
-      margin-left: -${p => math(`${p.theme.space.xxs} * 2 + ${p.theme.iconSizes.md}`)};
+      margin-left: -${p => math(`${p.theme.space.xxs} * 2 + ${p.theme.iconSizes.md} + ${p.theme.space.base}`)};
```
jzempel, line 27: Why were these margins modified? The addition of `theme.space.base` doesn't appear to be based on context logic.

jzempel (changes requested):
- the blue color lightening on focus for these custom semibold article components is not intended; please follow the original color design

Screenshot 2023-06-16 at 1 03 49 PM

The foreground color treatment should match...

Screenshot 2023-06-16 at 1 05 10 PM

- Missing ... https://github.com/zendeskgarden/website/blob/28b2a1d195c98091e277aa88ed503fcbecd17d74/src/layouts/Root/components/Footer.tsx#L15-L27

- The TOC border radius is broken (likely requires a custom treatment) ...

Screenshot 2023-06-16 at 1 07 28 PM

src/components/MarkdownProvider/components/Anchor.ts
Francois-Esquire: These margins were modified to account for `outline` overlap of the SVG. Compare with and without:
without
with

Francois-Esquire: Using the `theme.borderWidth.md` to position has the same effect as without (from above). The use of `margin-right` positions the headings in their current position while moving the anchor further out to avoid overlap of the focus ring.
border-width

jzempel: I see. Let's override (remove) the outline styling here, use `box-shadow` instead (as these are not wrap-able text strings; but more like buttons visually), and revert messing with the margining. The current layout isn't spaced as originally intended.

Francois-Esquire:
> * the blue color lightening on focus for these custom semibold article components is not intended; please follow the original color design
> 
> Screenshot 2023-06-16 at 1 03 49 PM
> The foreground color treatment should match...
> 
> Screenshot 2023-06-16 at 1 05 10 PM

Good catch @jzempel, this has been updated.

> * Missing ... https://github.com/zendeskgarden/website/blob/28b2a1d195c98091e277aa88ed503fcbecd17d74/src/layouts/Root/components/Footer.tsx#L15-L27

Can you specify what is missing?

> * The TOC border radius is broken (likely requires a custom treatment) ...
> 
> Screenshot 2023-06-16 at 1 07 28 PM

The TOC anchor has been updated and 4px `margin` was added to give the focus ring 1px of space. Open to what design deems appropriate in this case.

lucijanblagonic:
image
The focus outline is missing on the front page on some links.

image
The focus outline is missing on the top level pages with navigation.

image
**This looks good to me.** I like that the outline is not touching the border, looks much better IMHO.

Francois-Esquire:
> image The focus outline is missing on the front page on some links.
> 
> image The focus outline is missing on the top level pages with navigation.
> 

For these two with missing focus outline, we needed to include the latest `@zendeskgarden/css-bedrock`, which we now have!

focus-state-website-template-page
focus-state-homepage

> image **This looks good to me.** I like that the outline is not touching the border, looks much better IMHO.

Sounds good!

src/layouts/Home/components/News.tsx
```diff
@@ -8,12 +8,18 @@
...
-import { getColor } from '@zendeskgarden/react-theming';
+import { SELECTOR_FOCUS_VISIBLE, getColor } from '@zendeskgarden/react-theming';
 import { Grid, Row, Col } from '@zendeskgarden/react-grid';
 import { XXL } from '@zendeskgarden/react-typography';
 import { Anchor } from '@zendeskgarden/react-buttons';
 import MaxWidthLayout from 'layouts/MaxWidth';
 
+const StyledNewsAnchor = styled(Anchor)`
+  ${SELECTOR_FOCUS_VISIBLE} {
+    color: ${p => getColor('primaryHue', 700, p.theme)};
```
jzempel, line 19: The `:active` color change seen on the production website is missing when the news links have focus.

src/components/MarkdownProvider/components/Anchor.ts
```diff
@@ -5,37 +5,31 @@
...
+    margin-left: -${p => math(`${p.theme.space.xs} * 2 + ${p.theme.iconSizes.md} + ${p.theme.space.base}`)};
     border-radius: ${p => p.theme.borderRadii.md};
     padding: 0 ${p => p.theme.iconSizes.md};
     color: transparent;
 
     /* stylelint-disable selector-max-specificity */
     ${p => mediaQuery('down', 'md', p.theme)} {
-      margin-left: -${p => math(`${p.theme.space.xxs} * 2 + ${p.theme.iconSizes.md}`)};
+      margin-right: ${p => p.theme.space.base}px;
+      margin-left: -${p => math(`${p.theme.space.xxs} * 2 + ${p.theme.iconSizes.md} + ${p.theme.space.base}`)};
```
jzempel, line 28: Please see https://github.com/zendeskgarden/website/pull/496#discussion_r1235792121

src/layouts/Titled/components/TOC.tsx
```diff
@@ -52,10 +52,19 @@ export const TOCBlock: React.FC<{ data: IHeading[] } & HTMLAttributes<HTMLDivEle
 
 const StyledAnchor = styled(Anchor)<{ isCurrent: boolean }>`
   display: block;
+  position: relative;
   margin: ${p => p.theme.space.xxs} 0;
-  margin-left: ${p => p.isCurrent && `-${p.theme.borderWidths.sm}`};
-  border-left: ${p => p.isCurrent && `${p.theme.borders.sm} ${getColor('grey', 800, p.theme)}`};
+  margin-left: ${p => p.theme.space.xxs};
```
jzempel, line 57: Should be able to replace shorthand above with...

```suggestion
  margin: ${p => p.theme.space.xxs} 0 ${p => p.theme.space.xxs};
```

jzempel (changes requested):

src/components/MarkdownProvider/components/Anchor.ts
Francois-Esquire: The only part that is moving in the current layout is the anchor. Special attention was given to keep the title text in the original placement (the reason for the `margin-right` usage). Is the anchor shifting what's concerning or the title shifting?

jzempel: The outline rings here are larger than what we see with the current website's use of inset `box-shadow` and we know from past design passes that this is not desired for mobile. Please override to stick with the original `box-shadow: inset ...` plan.

Francois-Esquire: Just updated!

src/components/MarkdownProvider/components/Anchor.ts
```diff
@@ -5,22 +5,22 @@
...
 export const StyledAnchor = styled(Anchor).attrs(props => ({
   /* eslint-disable-next-line prefer-named-capture-group */
   isExternal: props.href && !/^(#|\/(?!\/))/u.test(props.href)
 }))`
+  ${({ theme }) =>
+    focusStyles({
+      theme,
+      inset: true,
+      styles: { outlineColor: 'transparent' }
+    })}
```
jzempel, line 22: Shouldn't this only apply to the `&.anchor.before` so that only the heading link icons are affected?

jzempel (approved):

PR #499, opened 2023-06-23
fix(layouts): update footer link outline color

Francois-Esquire:
## Description

Updates the website footer links focus indicator with compliant WCAG color contrast.

## Details

Before:
before
After:
after

## Checklist

### DIFF
--- a/src/layouts/Root/components/Footer.tsx
+++ b/src/layouts/Root/components/Footer.tsx
@@ -22,6 +22,10 @@ const StyledFooterItem = styled(Link)`
   }
 
   ${SELECTOR_FOCUS_VISIBLE} {
+    /*
+      updates the default outline-color from
+      css-bedrock for the darker footer background
+    */
     outline-color: currentcolor;
   }
 


src/layouts/Root/components/Footer.tsx
```diff
@@ -21,6 +21,14 @@ const StyledFooterItem = styled(Link)`
     color: inherit;
   }
 
+  ${SELECTOR_FOCUS_VISIBLE} {
+    /*
+      updates the default outline-color from
+      css-bedrock for the darker footer background
+    */
+    outline-color: currentcolor;
```
jzempel, line 29: This fix feels like it belongs in `css-components` in order to make the anchor reset more bullet-proof.

Francois-Esquire: Well, if we are in `:active` or `:hover` state, this would result in darker (blue-600+) focus indicators. Wouldn't this deviate from design intention?

lucijanblagonic (approved):
LGTM! This is the same treatment/approach as we have for links in our Global alert banners.

jzempel (approved):

PR #508, opened 2023-07-17
feat: new combobox component page

jzempel:
## Description

This component page is categorized under "Forms" and supports the new `Combobox` from the `@zendeskgarden/react-dropdowns.next` package. Example code was provided to demonstrate the expansive variants the `Combobox` can provide. Rather than building on one "grouped/filtered/autocomplete" default, the example code is structured to focus on individual capabilities that can then be combined by the consumer in a variety of ways. Please review and critique accordingly.

### Details

Note that the native "Select" page was also relocated under the "Forms" category. Eventually, when the new "Menu" is documented (at the navigation root) everything under "Dropdowns" will be deprecated and the section will go away with the v9 major release.

## Checklist


diana-oum (commented):
image
The warning content doesn't seem to fit the paradigm of the other types of validations because it's just describing the cactus whereas the error says it's not available. I think for warning, it should something like "Only 2 left in inventory". Also CD [does not recommend](https://www.figma.com/file/6g87L4FdKZTA3knt3Rsfdx/Garden?type=design&node-id=9911%3A43547&mode=design&t=eyD8dpqKEeRlrmon-1) "succcessfully" in a confirmation message -- it's not necessary because the user knows the request is accepted when we just state it.

jzempel:
> Do we need redirects for any of the legacy pages?

Ah good call, @geotrev – planned on it and slipped my mind. I added the missing `select-native` redirect.

lucijanblagonic:
Everything looks great, only a couple of nitpicks:

## [Multiselect tags](https://64b6ecf9adc6950435fa9243--zendeskgarden.netlify.app/components/combobox#multiselect-tags)

image

* I would add the arrow on the right, this section is talking about modifying tags. If the intent is to communicate you can search, we should have a search icon (but I think that makes it more complicated)
* I wouldn't use meta information for last names since the selected Tag displays it in a single row, without any difference in formatting

## [Select](https://64b6ecf9adc6950435fa9243--zendeskgarden.netlify.app/components/combobox#select-only)

Using `Space` to activate the listbox (as you [mentioned](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/#kbd_label) in the chat).

## [Media](https://64b6ecf9adc6950435fa9243--zendeskgarden.netlify.app/components/combobox#media)

image

I suggest we tweak this example if possible. Media communicates that designers can put in different elements next to text, and the focus will clear out those to allow for simple text input. I propose something simple as using a `green-200` `24px` circle with the emoji inside, placed on the left of the text. This will make the interaction more obvious as the user could just input the emoji in the input, and in that case it wouldn't be cleared as it behaves as text.

Overall, I like the structure, it's very informative and much easier to get your head around what this component does. 🙌

lucijanblagonic (approved):

Francois-Esquire (approved):
Looks solid.

jzempel:
@lucijanblagonic...

> I would add the arrow on the right

Part of what I need to do with these examples is demonstrate simplicity along with the ability to build up complexity. For the "Horticulturalists" I really want to focus on tag customization, so I left `isAutocomplete` (which requires a fair bit of regex filtering code) behavior off. When `isAutocomplete` is added, the chevron will appear along with click-to-dropdown functionality.

> I wouldn't use meta information for last names

Got it; I updated the tags to just show the name and not the meta.

> I suggest we tweak this example if possible

Brilliant, thanks for the feedback. I'll make these improvements.

PR #530, opened 2023-09-07
chore: upgrade react-components to 8.69.11

jzempel:
## Description

Upgrades `react-components` to 8.69.11 with patch fix for ID generation.


smritimadan (approved):
Deployment is looking good, labels and hints are wired up correctly on first and subsequent renders. 👍

geotrev (approved):

PR #533, opened 2023-10-04
feat(menu): adds menu page and marks legacy dropdown as deprecated

geotrev:
👉🏻 [Latest deploy preview](https://6526d50b0c96f50463818fd8--zendeskgarden.netlify.app/components/menu)

## Description

- Marks components in navigation as [deprecated] (`Dropdowns` > `Menu`)
- Marks Menu v.next as [new]. 
- Renames `/components/menu` to `/components/legacy-menu`
- Adds new `/components/menu` for Menu v.next
- Updates these pages that integrated legacy Dropdown/Menu, but now using v.next:
  - `SplitButton`
  - `InputGroup`

## Checklist


geotrev:
Currently waiting for a [functional change](https://github.com/zendeskgarden/react-containers/pull/604) from `react-containers` that will allow us to showcase consumers' ability to react to specific menu items, similar to the legacy menu examples.

src/examples/menu/Button.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <Row justifyContent="center">
+      <Col textAlign="center">
+        <Menu
+          button={props => (
+            <Button {...props}>
+              Menu
+              <Button.EndIcon isRotated={rotated}>
+                <ChevronIcon />
+              </Button.EndIcon>
```
jzempel, line 32: Dang. It just now occurs to me that our `Menu` component should be doing this out of the box. We might need to deliver an enhancement.

src/examples/menu/Nested.tsx
```diff
@@ -0,0 +1,68 @@
...
+      setItems(isNext ? NESTED_ITEMS : BASE_ITEMS);
+    } else if (isExpanded === false) {
+      setItems(BASE_ITEMS);
+    }
+  }, []);
+
+  return (
+    <Row justifyContent="center">
+      <Col textAlign="center">
+        <Menu button="Menu" onChange={handleChange}>
```
jzempel, line 48: I'm noticing interesting (bug?) mousing behavior where clicking on "Berry" and subsequently hovering over "Strawberry" doesn't change the styling until a different item is hovered. Something to consider for the `Menu` or container?

src/examples/menu/Size.tsx
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Col, Row } from '@zendeskgarden/react-grid';
+import { Menu, Item } from '@zendeskgarden/react-dropdowns.next';
+
+const Example = () => (
```
jzempel, line 12: Let's repeat the side-by-side from the original in order to drive home the sizing differentiation.

jzempel (changes requested):
To reinforce, I'm convinced the lack of a default button chevron is more of a critical miss on the base component since it simplifies UX that violates Garden's intended design pattern.

src/examples/menu/Button.tsx
geotrev: Hmmm, the legacy menu wasn't doing this so I presumed we would be the same here. But it's not difficult to implement, so definitely fair game to add in.

src/examples/menu/Size.tsx
geotrev: Should be updated now. :)

src/examples/menu/Nested.tsx
geotrev: Small update, looks like the behavior is correct in `containers` but broken in `components`. Woohoo? 🤦🏻

src/examples/menu/Button.tsx
geotrev: Will uptake new Menu version when released this week - I'll modify this example to demonstrate custom `button` another way so there's no conflict.

abbelco:
@geotrev Some notes on doc:
- Is Menu is the right place in the nav? Should we combine with other dropdowns under Forms? Or are we separating because there's no input in Menu
- "Menus help users choose items from a list." Is not really telling people what it is - what about something like: "A Menu is a dropdown component that contains a list of actions and supports keyboard navigation, sub-menus, and a wide variety of menu item types." Or something similar cc @jzempel 
- On the default button didn't we agree to the chevron?
- Under arrow - there's a delay in the arrow animating in - can we sync the timing of the menu and arrow at the same time? 
- Under grouped items the spacing looks off by a pixel with the check next to cactus - can we check that?
- The example under nested shows a user clicking on strawberry and then the title is fruits - shouldnt the title match the item the user clicked on?
- Placement: is this supposed to be centered this way in the middle or off to side? Probably a design question cc @lucijanblagonic

zendesk-rickschmoo:
1. I like the deprecated box at the top of the old Menu. I rec adding a box at the top of the new menu too to say that it is the new source of truth.
2. Why is Combobox under Forms but menu is not? Personally, I prefer at the top level as it seems fairly arbitrary what is under Forms and what isn't.

geotrev:
@abbelco Thanks for Qs and thoughts! Let me know if anything is unclear. Largely I copied over the [legacy Menu doc](https://garden.zendesk.com/components/menu) and spiced it up a little to showcase the newer API. 

> Is Menu is the right place in the nav? Should we combine with other dropdowns under Forms? Or are we separating because there's no input in Menu

Menus aren't form elements as they don't provide form functionality. They are the generic form of a dropdown. Not having an input is definitely the main aspect, you're right.

> "Menus help users choose items from a list." Is not really telling people what it is - what about something like: "A Menu is a dropdown component that contains a list of actions and supports keyboard navigation, sub-menus, and a wide variety of menu item types." Or something similar

Well, I never said I was a copy writer. ;) 

> On the default button didn't we agree to the chevron?

Yep. That is already merged over in `react-components` and will be merged added to this branch once it's released.

> Under arrow - there's a delay in the arrow animating in - can we sync the timing of the menu and arrow at the same time?

This is by design and matches the legacy menu animation. I would defer to @lucijanblagonic or @jzempel if the timing needs to be adjusted. 

> Under grouped items the spacing looks off by a pixel with the check next to cactus - can we check that?

Can you clarify what you mean here? If you're referring to the icon items, they're always indented. As far as I can see it matches the legacy menu / existing design.

Legacy menu:
Screenshot 2023-10-09 at 1 44 21 PM

Menu v.next:
Screenshot 2023-10-09 at 1 46 48 PM

> The example under nested shows a user clicking on strawberry and then the title is fruits - shouldnt the title match the item the user clicked on?

The title of the sub menu in this case matches the root menu's title, which is the label of the menu trigger. The legacy menu nested example is also this way, FWIW.

> Placement: is this supposed to be centered this way in the middle or off to side? Probably a design question

The `"top"` placement centers the menu - which is different than the legacy Menu placement example, which is `"top-start"`. I can revert back to that if it seems easier to understand/preferred.

geotrev:
Thanks @zendesk-rickschmoo !

> Why is Combobox under Forms but menu is not? Personally, I prefer at the top level as it seems fairly arbitrary what is under Forms and what isn't.

Menus are the abstract/base entity that Combobox extends from, but the latter supports forms inherently due to its input field. They are indeed separate constructs. I would defer to design on the ideal navigation structure - I haven't personally thought too deeply about it right here.

abbelco:
@geotrev @lucijanblagonic Under "Grouped Items" Cauliflower looks off in alignment by a pixel or two - can you confirm this?
image

abbelco:
@geotrev @lucijanblagonic should this be center? My expectation is off to side: 
image

jzempel:
> Under "Grouped Items" Cauliflower looks off in alignment by a pixel or two - can you confirm this?

@abbelco I think you're being thrown by the pixelated bowl of the "C". These words are exactly left aligned.

jzempel:
@zendesk-rickschmoo ...

> I rec adding a box at the top of the new menu too to say that it is the new source of truth.

We didn't do this with `Combobox` as we're not looking to produce extra upkeep. Let's revisit this post the PR.

> Why is Combobox under Forms but menu is not? Personally, I prefer at the top level as it seems fairly arbitrary what is under Forms and what isn't.

There is a general Garden design goal of flattening the nav so our consumers have clarity on the breadth of available components. We do want to retain certain categorization to establish clear groupings (regardless of `react-components` packages). We'll keep at least `Buttons` and `Forms` as helpful groupings (and probably some aspects of `Typography`). The `Menu` is categorically _not_ part of collecting form data and we would never want to hint that it should be. The important "Not for this" bullet list at the top of the page intends to make this clear.

src/examples/button/SplitButtonDanger.tsx
```diff
@@ -5,33 +5,34 @@
...
         <SplitButton>
           <Button isDanger>Harvest</Button>
-          <Dropdown
-            onStateChange={options =>
-              Object.hasOwn(options, 'isOpen') && setRotated(options.isOpen)
-            }
+          <Menu
+            button={props => (
+              <ChevronButton {...props} aria-label="other actions" isRotated={rotated} isDanger />
+            )}
```
jzempel, line 28: We'll be able to eradicate these with the upcoming `Menu` minor release, right?

jzempel (commented):
Should there be an example demonstrating anchor `Item` usage? I feel like that's a big win with this refactor that shouldn't go unnoticed.

src/examples/button/SplitButtonDanger.tsx
geotrev: Looks like we have two options here as far as recommended API usage.

Option 1:

The change currently in this PR.

Option 2:

```
const Example = () => (
  <Row>
    <Col textAlign="center">
      <SplitButton>
        <Button isDanger>Harvest</Button>
        <Menu button="" buttonProps={{ isDanger: true }} placement="bottom-end">
          <Item value="prune">Prune</Item>
          <Item value="water">Water</Item>
          <Item value="fertilize">Fertilize</Item>
        </Menu>
      </SplitButton>
    </Col>
  </Row>
);
```

---

The reason for the difference is that `button` is a required prop - but can be left as an empty string to fulfill its required typing, while minimizing the additional state logic. Option 1 feels weird but is possible & has some minimalist benefits.

src/pages/components/menu.mdx
```diff
@@ -48,102 +47,125 @@ components:
...
+import MediaCode from '!!raw-loader!../../examples/menu/Media.tsx';
 
-<CodeExample code={MenuMediaCode}>
-  <MenuMedia />
+<CodeExample code={MediaCode}>
+  <Media />
 </CodeExample>
 
-### Meta
+### Meta items
```
diana-oum, line 114: Could we make this "Meta text"? So it's conceptually consistent throughout the document.

geotrev: Thanks, pushed up that change.

src/examples/button/SplitButtonDanger.tsx
geotrev: Another fair point might be to remove required condition on `button` for this exact use-case.

jzempel: Ok, I'll look to you to go with whatever approach is cleanest.

jzempel (approved):

lucijanblagonic:
Thanks for addressing all the comments @geotrev great improvements already. Here are couple of my observations:

I don't get why the `Button` [example](https://6525a7c25f6eb225fe04fd77--zendeskgarden.netlify.app/components/menu#button) changes color/type from Outline to Primary? It's not something we would want to people to do in this context as even [Toggle buttons](https://garden.zendesk.com/components/toggle-button#type) always remain the same type. It also makes this weird icon transition glitch. My recommendation is to just use `isPrimary` in the example for custom menu trigger.

custom-menu-trigger

----

> @geotrev @lucijanblagonic should this be center? My expectation is off to side: image

I know that designers often find `placement` confusing (especially in placement) but in this case, I think the `top` placement as @geotrev added looks good and communicates that it's a custom setting. `start` and `end` placement look really weird when it comes to Menus that have a chevron pointing up/down.

----

Talking about animation, I was checking the [Tooltip modal](https://garden.zendesk.com/components/tooltip-modal) ([storybook](https://zendeskgarden.github.io/react-components/?path=/story/packages-modals-tooltipmodal--tooltip-modal&args=referenceElement:!null)), and something seems off. Was the animation changed recently, is it a bug? It starts and end abruptly, even worse when there are buttons inside.

tooltip-modal-animation

And some overall comments on transitions, I think this is something we should document better and perhaps rethink in some cases. I always expect the Chevron should be rotating faster, and not lagging behind the dropdown. But I don't want us to make any rush decisions here, just FYI. :)

jzempel:
> And some overall comments on transitions, I think this is something we should document better and perhaps rethink in some cases. I always expect the Chevron should be rotating faster, and not lagging behind the dropdown. But I don't want us to make any rush decisions here, just FYI. :)

Thanks @lucijanblagonic. Garden animations haven't had a thorough critique since initial design. Let's get a comprehensive review on the backlog and review outside the scope of this PR.

src/pages/components/legacy-menu.mdx
```diff
@@ -0,0 +1,290 @@
...
+Media elements add even more context to a Menu.
+
+import MenuMedia from '../../examples/dropdowns/menu/Media.tsx';
+import MenuMediaCode from '!!raw-loader!../../examples/dropdowns/menu/Media.tsx';
+
+<CodeExample code={MenuMediaCode}>
+  <MenuMedia />
+</CodeExample>
+
+### Meta
```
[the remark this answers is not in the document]
geotrev: (per separate convo, we're foregoing this change as it's on the legacy menu page)

src/examples/button/SplitButtonDanger.tsx
geotrev: I ended up leaving this code as-is. Menu is using `Button.IconEnd` which adds space when `button` is an empty string. It will be the same implementation as before, at least. :)

lucijanblagonic (approved):

diana-oum (approved):

PR #577, opened 2024-04-25
fix: use updated node IDs in save pattern assets

geotrev:
## Description

Adds new node IDs for save pattern page to fix the build.


ze-flo (approved):
🤞

jzempel (approved):
🎉

PR #582, opened 2024-05-06
fix(build): split figma v1/images API calls into chunks

### LINKED ISSUE #580: Figma REST API returning 400 error
## Overview

Garden extracts assets from a Figma file to populate images across several pages. This is primarily through the use of two APIs, the [files](https://www.figma.com/developers/api#get-files-endpoint) and [images](https://www.figma.com/developers/api#get-images-endpoint) endpoints.

We generalize the API call using a [function utility](https://github.com/zendeskgarden/website/blob/main/plugins/gatsby-source-garden-content/utils.ts#L58). Garden then hits the `files` and `images` endpoints during the build, first by getting the [frame node IDs](https://github.com/zendeskgarden/website/blob/main/content/figma/assets.yml), [here](https://github.com/zendeskgarden/website/blob/main/plugins/gatsby-source-garden-content/utils.ts#L98) (files) and [here](https://github.com/zendeskgarden/website/blob/main/plugins/gatsby-source-garden-content/utils.ts#L117) (images), respectively.

## Expectations

The endpoints consistently return the associated file/image assets within the 55 second timeout allowed by the REST API, as they have been for the past year.

## Reality

The API has suddenly began returning 400 errors. This happened before, about a week ago, and the "answer" at the time appeared to be deleting and recreating frames within our assets Figma file. I assume the same issue is continuing here.

To be clear, it _appears_ that specific frames are buggy/erroneous, delaying the API call and resulting an error state. Previously, each of the API calls would take ~20 seconds each, well within the timeout.

## Steps to Reproduce

1. Clone the repo and install the dependencies (`npm i`). You should be on `main`.
2. The build should run as a `postinstall` script – you'll see this Figma 400 error:
    Screenshot 2024-04-30 at 3 04 06 PM

To make the build "work" (and test that this is a Figma issue):

1. Run `rm -rf public .cache` to clear build artifacts.
2. Delete the folder `src/pages/patterns` and delete all [asset entries](https://github.com/zendeskgarden/website/blob/main/content/figma/assets.yml) whose node name begins with `patterns-`.
3. Run `npm run build` and notice the build passes without errors.

### Fine Print

- URL: `https://api.figma.com/v1/images/HifWgSrdeTlMMXUrcnrAAH?ids=*`
- Version: Latest
- Environment: Node / Figma API

geotrev:
## Description

Fixes #580 (for full context of the root issue being solved)

Splits the `v1/images` API call into smaller chunks to prevent timeout errors.

## Detail

Figma's REST API has a tendency to timeout if you request too many node IDs on their `v1/images` endpoint.

Their recommendation is to do multiple small requests to ensure all assets are returned quickly without hitting the timeout.

In our code, the solution then is to chunk the node ID list into a few small batches and run a fetch request on each, then merge the responses. Since `fetch` is async, a `for await` loop is leveraged to safely resolve each request sequentially.

### DIFF
--- a/plugins/gatsby-source-garden-content/utils.ts
+++ b/plugins/gatsby-source-garden-content/utils.ts
@@ -9,6 +9,7 @@ import { readFileSync } from 'fs';
 import { resolve, join } from 'path';
 import { BinaryToTextEncoding, createHash } from 'crypto';
 import { parse } from 'yaml';
+import { chunk } from 'lodash';
 
 export function hashFileContent(
   content: string,
@@ -113,10 +114,20 @@ export async function fetchFigmaImages({
   nodeIds: string[];
   scale: number;
 }) {
-  const idsParam = nodeIds.join(',');
-  const { images } = await fetchFigmaAPI(`images/${fileId}?ids=${idsParam}&scale=${scale}`, {
-    figmaApiToken
-  });
+  const chunks = chunk(nodeIds, 40);
+  let images: Record<string, string> = {};
+
+  for await (const value of chunks) {
+    const idsParam = value.join(',');
+    const { images: _images } = await fetchFigmaAPI(
+      `images/${fileId}?ids=${idsParam}&scale=${scale}`,
+      {
+        figmaApiToken
+      }
+    );
+
+    images = { ...images, ..._images };
+  }
 
   return { images };
 }


plugins/gatsby-source-garden-content/utils.ts
```diff
@@ -113,10 +114,20 @@ export async function fetchFigmaImages({
...
   scale: number;
 }) {
-  const idsParam = nodeIds.join(',');
-  const { images } = await fetchFigmaAPI(`images/${fileId}?ids=${idsParam}&scale=${scale}`, {
-    figmaApiToken
-  });
+  const chunks = chunk(nodeIds, 40);
+  let images: Record<string, string> = {};
+
+  for await (const value of chunks) {
```
ze-flo, line 120: Maybe a little comment to explain why we have to follow this approach?

geotrev: Apologies, I'll add a description shortly. I meant to keep this in draft until I got more answers from Figma's tech support. 😅

jzempel (approved):
looking good 🔍

PR #598, opened 2024-07-30
chore: prepare theming upgrade with getColorV8

geotrev:
## Description

Prepare for the v9 (specifically theming) upgrade by moving to `getColorV8`. We will need to eventually mark the utilities section for `getColorV8` deprecated, and add in a new section for `getColor` for V9 theming.

### DIFF
--- a/src/examples/utilities/components.ts
+++ b/src/examples/utilities/components.ts
@@ -100,7 +100,7 @@ const components = [
     }
   },
   {
-    name: 'getColor',
+    name: 'getColorV8',
     props: {
       hue: {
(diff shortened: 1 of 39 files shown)


src/examples/utilities/components.ts
```diff
@@ -100,7 +100,7 @@ const components = [
     }
   },
   {
-    name: 'getColor',
+    name: 'getColorV8',
```
ze-flo, line 103: TODO: rename to getColor with V9

ze-flo (approved):

PR #603, opened 2024-08-01
refactor: upgrade MardownProvider for `v9`

ze-flo:
## Description
Small PR to update MarkdownProvider and subcomponents for `V9`. Light / dark theming changes to be addressed later.

## Detail
- Sets the default ColorPicker color to the default offset for light mode
- Point CodeSandbox Garden deps to `next` tag


src/components/MarkdownProvider/components/CodeExample/index.tsx
```diff
@@ -119,7 +119,7 @@ export const CodeExample: React.FC<ICodeExampleProps> = ({ children, code }) =>
             <DirectionRtlStroke />
           </ToggleIconButton>
         </Tooltip>
-        <ColorPickerDialog color={color} onChange={setColor}>
+        <ColorPickerDialog color={color} onChange={setColor} placement="bottom-end">
```
ze-flo, line 122: Improves the `ColorPickerDialog` placement on narrow screens

Screenshot 2024-08-01 at 10 04 50 AM

ze-flo:
Verified that the `generateColorScale` was called when picking a custom primary color.

Screenshot 2024-08-01 at 10 03 17 AM

src/components/MarkdownProvider/components/CodeExample/index.tsx
geotrev: Interesting – I would have expected TooltipModal to automatically reposition the dialog here since it's value is defaulting to `auto`. 

https://github.com/zendeskgarden/react-components/blob/next/packages/modals/src/elements/TooltipModal/TooltipModal.tsx#L182

Was that not the case here?

ze-flo: It's set to `bottom-start` by default. In our case, that doesn't work well because of the picker location.

https://github.com/zendeskgarden/react-components/blob/3da9488a98d57c4150b0308aa67b9f561c4ef929/packages/colorpickers/src/elements/ColorSwatchDialog/index.tsx#L200-L204

geotrev (approved):

src/components/MarkdownProvider/components/CodeExample/index.tsx
geotrev: That looks like the color picker dialog in the screen shot, is that what we're talking about? (You linked to color swatch dialog).

geotrev: Oh I see now - both use bottom-start by default. 😅

In that case it looks good. I guess I was wondering if "auto" would achieve the desired result and be less specific to a single position.

ze-flo: Ooops, my bad.😓  That's what I meant. Both have the same `bottom-start` default.

PR #619, opened 2024-08-12
feat(components): update draggable to v9

geotrev:
## Description

**Package migration changes:**
- Updates `react-drag-drop` to `react-draggable` (v9)

**Other updates:**
- Updates examples 
  - Use `Grid` subcomponent properties
  - Apply custom colors to support dark mode (headings in `DraggableList` and `Dropzone` examples)

https://github.com/user-attachments/assets/83d92bf8-cc34-4559-a7f2-09450f9df511

Notes:
- No changes to MDX file(s)

## Checklist

- ~~:ok_hand: website updates are Garden Designer approved (add the designer as a reviewer)~~
- ~~:black_nib: copy updates are approved (add the content strategist as a reviewer)~~
- ~~:link: considered opportunities for adding cross-reference URLs (grep for keywords)~~
- ~~:wheelchair: analyzed via [axe](https://www.deque.com/axe/) and evaluated using VoiceOver~~
- ~~:memo: tested in Chrome, Firefox, Safari, Edge~~

### DIFF
--- a/src/examples/draggable/DraggableContent.tsx
+++ b/src/examples/draggable/DraggableContent.tsx
@@ -7,8 +7,8 @@
 
 import React from 'react';
 import { PALETTE, getColor } from '@zendeskgarden/react-theming';
-import { Row, Col } from '@zendeskgarden/react-grid';
-import { Draggable } from '@zendeskgarden/react-drag-drop';
+import { Grid } from '@zendeskgarden/react-grid';
+import { Draggable } from '@zendeskgarden/react-draggable';
 import { IconButton } from '@zendeskgarden/react-buttons';
 import { Tooltip } from '@zendeskgarden/react-tooltips';
 import { Span } from '@zendeskgarden/react-typography';
@@ -23,8 +23,8 @@ const StyledDecorator = styled.div`
 `;
 
 const Example = () => (
-  <Row justifyContent="center">
-    <Col sm={7}>
+  <Grid.Row justifyContent="center">
+    <Grid.Col sm={7}>
       <Draggable isCompact>
         <Draggable.Grip />
         <StyledDecorator>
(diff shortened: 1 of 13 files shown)


ze-flo (approved):
Looking good!

src/examples/draggable/DraggableListDefault.tsx
```diff
@@ -19,12 +20,14 @@ const items = [
 ];
 
 const StyledHeading = styled(MD)`
+  transition: color 0.2s ease-in-out;
   margin-bottom: ${p => p.theme.space.xs};
+  color: ${p => getColor({ variable: 'foreground.default', theme: p.theme })};
```
jzempel, line 25: Ideally, this will utilize Garden's `<Span>` with a semantic `hue` prop when that functionality becomes available (currently WIP) in order to demonstrate first-class Garden usage. This is also a good reminder for the `Span` to receive a `transition` property along with the new `hue` functionality. With that in mind, I wonder if we should just remove the added `transition` CSS from this PR?

src/examples/draggable/DraggableContent.tsx
```diff
@@ -23,8 +23,8 @@ const StyledDecorator = styled.div`
 `;
 
 const Example = () => (
-  <Row justifyContent="center">
-    <Col sm={7}>
+  <Grid.Row justifyContent="center">
+    <Grid.Col sm={7}>
```
jzempel, line 27: Should the "Oranges, mandarins, limes, and the like" `<Span>` and `StyledDecorator` receive `getColor({ variable: 'foreground.subtle' })` treatment? Or are we waiting for updated `<Span hue="foreground.subtle">` here?

src/examples/draggable/DraggableListDefault.tsx
geotrev: Agreed re: removing transition if that comes with Span / typography components. 💯

src/examples/draggable/DraggableContent.tsx
geotrev: `Draggable` brings along `foreground.default`, otherwise this woul definitely need the additional color support (like the other one you mentioned).

jzempel (approved):

PR #621, opened 2024-08-12
refactor(component pages): reference updated subComponents' `displayName`

ze-flo:
## Description
References updated subComponents' `displayName` to pull `props` info.

## Detail
- Bumps `next` @zendeskgarden deps to latest `next` release
- Bumps `react-components` submodule to latest `next` tag
- Updates MDX files with subcomponents where necessary
- References updated subComponents' `displayName`

## Checklist

- [ ] ~~:ok_hand: website updates are Garden Designer approved (add the designer as a reviewer)~~
- [ ] ~~:black_nib: copy updates are approved (add the content strategist as a reviewer)~~
- [ ] ~~:link: considered opportunities for adding cross-reference URLs (grep for keywords)~~
- [ ] ~~:wheelchair: analyzed via [axe](https://www.deque.com/axe/) and evaluated using VoiceOver~~
- [ ] ~~:memo: tested in Chrome, Firefox, Safari, Edge~~

PR #623, opened 2024-08-13
chore: removes ESLint `--max-warning` flag from lint-staged config

ze-flo:
## Description

Temporarily removes ESLint --max-warning flag from lint-staged config (omission from refactor(component pages): reference updated subComponents' displayName #621).


geotrev (approved):
Good call on a separate PR. 🚀

PR #624, opened 2024-08-13
feat(components): update `tables` to v9 

ze-flo:
## Description

Updates `react-tables` to v9.

## Details
- Refactors examples to use `Table`'s subComponent properties

## V9 Migration checklist
- [x] Bump @zendeskgarden dependency
- [ ] ~~Update imports source~~
- [x] `Table` - Update examples to subcomponent properties 
- [x] `Field` - Update examples to subcomponent properties
- [ ] ~~`Grid` - Update examples to subcomponent properties~~
- [ ] ~~Update API breaking changes~~
- [x] MDX - Update docs to reflect migration changes (handled [here](https://github.com/zendeskgarden/website/pull/621/files#diff-e8f7f0a27eac30f4e10609fb83487f1571ec812bb53db2fe9b748bb97a14b4aeR205-R223))
- [x] MDX - Update to subcomponent properties (handled [here](https://github.com/zendeskgarden/website/pull/621/files#diff-e8f7f0a27eac30f4e10609fb83487f1571ec812bb53db2fe9b748bb97a14b4aeR236-R309))

Screenshot 2024-08-13 at 10 53 06 AM
Screenshot 2024-08-13 at 10 52 43 AM

## Checklist

- [ ] ~~:wheelchair: analyzed via [axe](https://www.deque.com/axe/) and evaluated using VoiceOver~~
- [ ] ~~:memo: tested in Chrome, Firefox, Safari, Edge~~


src/examples/table/TableVerticalScrolling.tsx
```diff
@@ -18,15 +18,15 @@ const rowData = Array.from(Array(10)).map((row, index) => ({
   soil: 'Well draining'
 }));
 
-const StyledHeaderCell = styled(HeaderCell)`
+const StyledHeaderCell = styled(Table.HeaderCell)`
   box-shadow: inset 0 -1px 0 ${props => getColorV8('neutralHue', 300, props.theme)};
```
geotrev, line 22: Shoudl this swap over to `getColor`? Not sure if there's other custom colors in here. Also not sure if this can take advantage of the new `shadow` variables. 😮

src/examples/table/TableDraggable.tsx
```diff
@@ -8,11 +8,11 @@
 import React, { PropsWithChildren, useCallback, useEffect, useState } from 'react';
 import styled from 'styled-components';
 import { DragDropContext, Droppable, Draggable, DropResult } from 'react-beautiful-dnd';
-import { Table, Row, Cell, Head, HeaderCell, HeaderRow, Body } from '@zendeskgarden/react-tables';
+import { Table } from '@zendeskgarden/react-tables';
 import { ReactComponent as GripIcon } from '@zendeskgarden/svg-icons/src/12/grip.svg';
 import { getColorV8 } from '@zendeskgarden/react-theming';
 
-const DraggableRow = styled(Row)<{ isDraggingOver: boolean }>`
```
geotrev, line 15: Seeing some `getColorV8` in here as well.

src/examples/table/TableDraggable.tsx
```diff
@@ -104,7 +104,7 @@ class DraggableCell extends React.Component<IDraggableCellProps> {
   }
```
geotrev, line 104: The use of `primaryHue` for the draggable grip feels like it needs some design review. I'm sure they'll come across it but just wanted to leave a comment about it here for posterity. 🧐

src/examples/table/TableScroll.tsx
```diff
@@ -24,22 +24,22 @@ const rowData: IRow[] = Array.from(Array(100)).map((row, index) => ({
 const Example = () => (
   <div style={{ maxHeight: 500, overflow: 'auto' }}>
```
geotrev, line 25: This needs `color-scheme` that follows light/dark. In `react-components`, we used `only light` and `only dark`, respectively. :)

geotrev (changes requested):
I noticed some colors lacking `transition` for certain properties:

- `Head` background color
- `Table.Row` border

Feels like good follow-ons (not blocking this PR) / work for when we're doing QA passes. Should we write them down somewhere? Perhaps in a ticket? 👨🏻‍🍳

src/examples/table/TableDraggable.tsx
geotrev: I suspect they'll suggest we update it to `foreground.subtle` like in `Draggable.Grip`, but maybe there's a reason it's blue here.

src/examples/table/TableVerticalScrolling.tsx
```diff
@@ -18,15 +18,17 @@ const rowData = Array.from(Array(10)).map((row, index) => ({
   soil: 'Well draining'
 }));
 
-const StyledHeaderCell = styled(HeaderCell)`
-  box-shadow: inset 0 -1px 0 ${props => getColorV8('neutralHue', 300, props.theme)};
```
ze-flo, line 22: Removed since v9 `StyledHead` component already [sets](https://github.com/zendeskgarden/react-components/blob/next/packages/tables/src/styled/StyledHead.ts#L26) the `box-shadow` correctly.

src/examples/table/TableDraggable.tsx
```diff
@@ -24,11 +24,7 @@ const DraggableRow = styled(Row)<{ isDraggingOver: boolean }>`
 `;
 
 const DraggableContainer = styled.div`
-  color: ${props => getColorV8('primaryHue', 600, props.theme)};
-
-  &:hover {
-    color: ${props => getColorV8('primaryHue', 700, props.theme)};
-  }
+  color: ${p => getColor({ variable: 'foreground.subtle', theme: p.theme })};
```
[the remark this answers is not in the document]
geotrev: Nice. I also raised a question to design in Slack so let's see what they say. 🧐

src/examples/table/TableDraggable.tsx
jzempel, line 1: Looks like this example is broken in CodeSandbox. But it seems to have been broken here too https://garden.zendesk.com/components/table#draggable. Might not be worth investigation, as our preference would be to move this away from `react-beautiful-dnd`.

src/examples/table/TableDraggable.tsx
jzempel: Agreed. I don't think there is substantial design reasoning behind the primary blue. It was originally applied as a one-off in this example. The Garden designed `Draggable.Grip` should be considered to be the guiding principle.

src/examples/table/TableVerticalScrolling.tsx
```diff
@@ -18,15 +18,17 @@ const rowData = Array.from(Array(10)).map((row, index) => ({
...
+const StyledContainer = styled.div`
+  max-height: 400px;
+  overflow-x: auto;
+  color-scheme: only ${p => p.theme.colors.base};
```
jzempel, line 24: All of these `color-scheme` CSS properties is making it apparent that the `<Table>` component itself should have this CSS and should be removed from all of these examples. Doing so will apply nicely to all of the product usage of Garden scrolling tables.

jzempel (changes requested):

src/examples/table/TableVerticalScrolling.tsx
geotrev: This is a challenging scenario because Garden `Table` has no scroll by default (same as any other basic HTML table, also the reason W3C's [only article](https://www.w3schools.com/howto/howto_css_table_responsive.asp) suggests using a container, same as we do here).

We would need to provide some additional API to apply `overflow` CSS to `Table`/`Table.Body` ([one](https://www.geeksforgeeks.org/how-to-create-table-with-100-width-with-vertical-scroll-inside-table-body-in-html/), [two](https://vembarrajan.medium.com/html-css-tricks-scroll-able-table-body-tbody-d23182ae0fbc)).

So, all that said, perhaps we can do both: provide scrolling via `Table` prop + move `color-scheme` into `CodeExample.tsx` (which would apply to all website examples).

src/examples/table/TableDraggable.tsx
ze-flo: I briefly investigated but the root-cause of the error isn't straightforward. Since we want to migrate away from `react-beautiful-dnd`, should I refactor using [DnD Kit](https://dndkit.com/)?

ze-flo: [Refactored](https://github.com/zendeskgarden/website/pull/624/commits/9554a246ae25bd12c71a47aa9ca0e6a14143ab02) using [@dnd-kit/sortable](https://docs.dndkit.com/presets/sortable). Fantastic library and now it works flawlessly on CodeSandbox! 💯 

I still have to figure out why removing `react-beautiful-dnd` causes the build to fail. 🤔 🤷‍♂️ 

https://app.circleci.com/pipelines/github/zendeskgarden/website/4027/workflows/e3b32f08-2ebf-4a21-a598-2ce13fe77d4c/jobs/8606

src/examples/table/TableVerticalScrolling.tsx
jzempel: Oh, you're totally right @geotrev – there's not a quality addition to the `Table` component 🤦

src/examples/table/TableDraggable.tsx
jzempel, line 1: 🎗️ to update the [description](https://66bdaee0196ce983c15b2761--zendeskgarden.netlify.app/components/table#draggable) associated with this example, as it currently refers to `react-beautiful-dnd`

jzempel (changes requested):

src/examples/table/TableDraggable.tsx
```diff
@@ -5,109 +5,37 @@
...
+} from '@dnd-kit/sortable';
+import { CSS } from '@dnd-kit/utilities';
 
 const DraggableContainer = styled.div`
-  color: ${props => getColorV8('primaryHue', 600, props.theme)};
-
-  &:hover {
-    color: ${props => getColorV8('primaryHue', 700, props.theme)};
-  }
+  cursor: grab;
```
jzempel, line 32: Please update the cursor to `grabbing` on drag-drop. See https://garden.zendesk.com/components/table#draggable for details

jzempel: It also looks like there's missing selected (light blue background) styling for the keyboard `<space>` action.

ze-flo: Fixed in [7f8796c](https://github.com/zendeskgarden/website/pull/624/commits/7f8796cd17337d952310a0c47efa9381c78801fd) ✅ 

Screenshot 2024-08-15 at 12 11 53 PM (2)

src/examples/table/TableDraggable.tsx
ze-flo: I removed `react-beautiful-dnd`, regenerated `package-lock.json`, removed `./node_modules`, and successfully ran `npm i`. 

When running `npm run build`, the previous [error](https://app.circleci.com/pipelines/github/zendeskgarden/website/4027/workflows/e3b32f08-2ebf-4a21-a598-2ce13fe77d4c/jobs/8606) `error Type with name "GardenNavigation" does not exists` no longer appears. However, a new error has emerged:

```
Error: Cannot find module 'ajv/dist/compile/codegen'
```

The error above is related to conflicts arising from having multiple versions of `ajv` as transitive deps. 

I propose handling the removal of `react-beautiful-dnd` in a follow-up / clean-up PR 🧼  to avoid blocking current migrations. 

Any objections?

src/examples/table/TableDraggable.tsx
```diff
@@ -86,11 +87,23 @@ const SortableRow = ({ item }: { item: IItem }) => {
...
+            })
+          : undefined
       }}
     >
       <Table.Cell>
-        <DraggableContainer {...attributes} {...listeners}>
+        <DraggableContainer
+          {...attributes}
+          {...listeners}
+          style={isDragging ? { cursor: 'grabbing' } : undefined}
```
jzempel, line 105: There seems to be a problem where the `grabbing` cursor reverts to default on drag. Can this be fixed?


https://github.com/user-attachments/assets/70ada191-91f7-47c1-88e8-6c19cede50ca

src/examples/table/TableDraggable.tsx
```diff
@@ -5,109 +5,37 @@
...
-  setRef(ref: HTMLTableCellElement | null) {
-    this.ref = ref;
-  }
-
-  render() {
-    return <Cell ref={this.setRef}>{this.props.children}</Cell>;
-  }
-}
-
 interface IItem {
```
jzempel, line 39:
```suggestion

interface IItem {
```

[nit] space styling

src/examples/table/TableDraggable.tsx
ze-flo: After hours of troubleshooting, I still wasn't able to pinpoint the cause of the weird cursor behavior. However, I did find a solution that improves the UX. Now, the cursor occasionally switches from `grabbing` to `grab`.

https://github.com/user-attachments/assets/76a08645-477c-4e2a-9123-0b1f1f2768e7

I hope this is an acceptable compromise.

jzempel: Sounds like a `dnd-kit` problem – probably not worth pouring more time into. LGTM

geotrev: Very weird. I dont think i've ever seen this problem... wonder if it exists in the dnd kit examples for draggable components.

geotrev (approved):

PR #629, opened 2024-08-14
feat(components): updates pagination to v9

geotrev:
## Description

Updates docs to support v9.

## Detail

- [x] Updates examples with [migration guide](https://github.com/zendeskgarden/react-components/blob/next/docs/migration.md) changes
- [x] Updates Grid subcomponent properties
- [x] Updates MDX files

## Checklist

- ~~:ok_hand: website updates are Garden Designer approved (add the designer as a reviewer)~~
- ~~:black_nib: copy updates are approved (add the content strategist as a reviewer)~~
- ~~:link: considered opportunities for adding cross-reference URLs (grep for keywords)~~
- ~~:wheelchair: analyzed via [axe](https://www.deque.com/axe/) and evaluated using VoiceOver~~
- ~~:memo: tested in Chrome, Firefox, Safari, Edge~~

### DIFF
--- a/src/examples/pagination/OffsetPadding.tsx
+++ b/src/examples/pagination/OffsetPadding.tsx
@@ -6,12 +6,7 @@
  */
 
 import React, { useState } from 'react';
-import styled from 'styled-components';
 import { OffsetPagination } from '@zendeskgarden/react-pagination';
-
-const StyledSpacer = styled.div`
-  height: ${p => p.theme.space.xl};
-`;
 
 const Example = () => {
   const [page, setPage] = useState(10);
@@ -19,9 +14,13 @@ const Example = () => {
   return (
     <nav aria-label="pagination example padding">
       <OffsetPagination totalPages={20} pagePadding={0} currentPage={page} onChange={setPage} />
-
-      <StyledSpacer />
-      <OffsetPagination totalPages={20} pagePadding={2} currentPage={page} onChange={setPage} />
+      <OffsetPagination
+        totalPages={20}
+        pagePadding={2}
+        currentPage={page}
+        onChange={setPage}
+        style={{ marginTop: '16px' }}
+      />
     </nav>
   );
 };
--- a/src/examples/pagination/OffsetPadding.tsx
+++ b/src/examples/pagination/OffsetPadding.tsx
@@ -6,21 +6,22 @@
  */
 
 import React, { useState } from 'react';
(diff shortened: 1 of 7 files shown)


src/pages/components/pagination.mdx
```diff
@@ -41,6 +41,18 @@ import CursorDefaultCode from '!!raw-loader!../../examples/pagination/CursorDefa
...
+Offset-based pagination allows users to see the total number of pages and jump
+to a specific position within the set.
+
+import OffsetDefault from '../../examples/pagination/OffsetDefault.tsx';
+import OffsetDefaultCode from '!!raw-loader!../../examples/pagination/OffsetDefault.tsx';
+
+<CodeExample code={OffsetDefaultCode}>
+  <OffsetDefault />
+</CodeExample>
+
```
geotrev, line 55: Minor change since the next two sections follow configurations of this component. If that seems incorrect we can reverse course.

src/examples/pagination/OffsetPadding.tsx
```diff
@@ -6,21 +6,22 @@
...
-        totalPages={20}
-        pagePadding={2}
-        currentPage={page}
-        onChange={setPage}
-        style={{ marginTop: '16px' }}
-      />
+      <OffsetPagination totalPages={20} pagePadding={0} currentPage={page} onChange={setPage} />
+
+      <StyledSpacer />
+      <OffsetPagination totalPages={20} pagePadding={2} currentPage={page} onChange={setPage} />
```
jzempel, line 24: Let's leave the `style` as-is with just the rename. Since two `<OffsetPagination>` components won't stack IRL, we don't need to demonstrate theming concerns with this example.

src/pages/components/pagination.mdx
jzempel: Hmm, originally component pages were structured to alphabetize the examples. Although that approach has been challenged (with Draggable, for example) my instinct is to just leave as-is until the website's content design receives a more thorough overhaul.

jzempel (changes requested):

geotrev:
I also noticed the new `labels` prop is shown as a `func` type (which based on the type check logic, is "correct"), even though it's an object. I suppose it's not the end of the world, but I wonder if we could revise this in some way to capture the correct type (e.g., `Record<...>`)

jzempel (approved):

PR #630, opened 2024-08-14
fix(build): reduce figma fetch size

geotrev:
🚧 🔧 Tinkering with Figma fetch requests to fix another suite of failing builds


ze-flo (approved):
🤞

jzempel (approved):

PR #626, opened 2024-08-13
feat(components): update `tabs` to v9

ze-flo:
## Description

Updates `react-tabs` to v9.

## Details
- Refactors examples to use `Tabs`' subComponent properties
- Set `CodeExample`’s text color to `foreground.default` var

## V9 Migration checklist
- [x] Bump @zendeskgarden dependency
- [ ] ~~Update imports source~~
- [x] `Tabs` - Update examples to subcomponent properties 
- [ ] ~~`Field` - Update examples to subcomponent properties~~
- [ ] ~~`Grid` - Update examples to subcomponent properties~~
- [ ] ~~Update API breaking changes~~
- [ ] ~~MDX - Update docs to reflect migration changes~~
- [x] MDX - Update to subcomponent properties (handled [here](https://github.com/zendeskgarden/website/pull/621/files#diff-719cdc2fa3f81eeff582ab2efde23b365b3b521b3619c1cbfc1e2aff0db79e3dR77-R99))

Screenshot 2024-08-13 at 11 34 35 AM
Screenshot 2024-08-13 at 11 34 51 AM

## Checklist

- [ ] ~~:ok_hand: website updates are Garden Designer approved (add the designer as a reviewer)~~
- [ ] ~~:black_nib: copy updates are approved (add the content strategist as a reviewer)~~
- [ ] ~~:link: considered opportunities for adding cross-reference URLs (grep for keywords)~~
- [ ] ~~:wheelchair: analyzed via [axe](https://www.deque.com/axe/) and evaluated using VoiceOver~~
- [ ] ~~:memo: tested in Chrome, Firefox, Safari, Edge~~

PR #631, opened 2024-08-14
fix(draggable): removes extraneous foreground colors in examples

geotrev:
Removing the extra foreground definitions given the central foreground update in #626


jzempel (approved):
nice 🧹

PR #636, opened 2024-08-22
refactor: migrate deprecated subComponent exports to subComponent properties

ze-flo:
## Description

Running ESLint with the `eslint-plugin-deprecation` plugin revealed instances of deprecated subComponent exports.


jzempel (approved):
Nice. Is the disabled deprecations rule ready to be removed? And what about the `--max-warnings 0` eslint flag?

geotrev (approved):
Nice catches!

ze-flo:
> Nice. Is the disabled deprecations rule ready to be removed? And what about the `--max-warnings 0` eslint flag?

The disabled deprecations rule should stay active and we should continue with `--max-warnings 0` until we replace all `getColorV8`.

PR #644, opened 2024-09-06
feat(combobox): add nested multiselect example

jzempel:
## Description

The new controlled example considers the following:
- in place `<Option>` swapping based on key next/previous values
- ~usage of root `<Option>` icons to indicate sub-option selection~ (TBD in v9)
- always rendering all selected options, including `<Option isHidden>` for accessibility
- a listbox collapse effect to restore root `<Option>` state
- controlling `activeIndex` for improved "skip past sub-header option" ergonomics

Screenshot 2024-09-06 at 11 05 41 AM

## Detail

Also adds similar enhancements to the existing nested (single select) example.

## Checklist


src/examples/combobox/NestedMultiselect.tsx
```diff
@@ -0,0 +1,185 @@
...
+interface IOption extends IOptionProps {
+  value: string;
+}
+interface IOptGroup extends IOptGroupProps {
+  options: IOption[];
+}
+
+type Options = (IOption | IOptGroup)[];
+
+type SubOptions = Record<any, Options>;
```
ze-flo, line 32: [nit]

```suggestion
type SubOptions = Record<string, Options>;
```

src/examples/combobox/NestedMultiselect.tsx
```diff
@@ -0,0 +1,185 @@
...
+        const subValues = option.options.map(subOption => subOption.value);
+
+        _values.push(...subValues);
+      }
+
+      return _values;
+    }, []);
+    const missingValues = state.selectionValue.filter(value => !values.includes(value));
+
+    return missingValues.map(value => <Option key={value as string} isHidden value={value} />);
```
ze-flo, line 143: Reminder: the casting `key={value as string}` is redundant with v9.

ze-flo (approved):
Beautiful work! ✨

src/examples/combobox/NestedMultiselect.tsx
jzempel: @ze-flo we'll get these type enhancements dialed-in for v9, where `OptionValue` is only valid for `string`. In order to let the `OptionValue = string | object` flow through the example, this has to be `any` 😞

src/examples/combobox/NestedMultiselect.tsx
jzempel: Yep, the v9 example is going to look cleaner ✨

geotrev (approved):
Non-blocking question because this would (presumably) be a concern on the container side: should an `OptGroup` with nested options respond to arrow right (or left, in RTL) to dig into nested options (similar to nested menu items)?

ze-flo:
To avoid merge conflicts, I'll handle the Renovate + Dependabot PRs once this one merges.

jzempel:
> To avoid merge conflicts, I'll handle the Renovate + Dependabot PRs once this one merges.

@ze-flo we're deferring all of the bot updates until after `next` -> `main`.

jzempel:
> Non-blocking question because this would (presumably) be a concern on the container side: should an `OptGroup` with nested options respond to arrow right (or left, in RTL) to dig into nested options (similar to nested menu items)?

Yeah, we should explore the container enhancement at some point. I think the question is how things will work with/against `downshift` :shrug:

steue (approved):
LGTM so far. looking forward to our improvement for the top level multi selection indicator

jzempel:
> Just noticed there is a small input bug when trying to backspace, the first character is never deleted.
> 
>  Screen.Recording.2024-09-11.at.9.28.40.AM.mov

Thanks, @geotrev. Those or conditions never quite work out like they look on paper 😜. Changed `changes.inputValue || state.inpuValue` to `changes.inputValue === undefined ? state.inputValue : changes.inputValue`

PR #652, opened 2024-10-02
fix: update `CodeSandbox` parameters to import v9 Garden dependencies

ze-flo:
## Description

Updates `CodeSandbox` parameters to import v9 Garden dependencies


jzempel (changes requested):
I've added a couple of suggestions that would be good to nab while we're here. 🙏

jzempel (approved):

PR #658, opened 2024-10-24
fix(docsearch): add trailing slashes to sitemap

jzempel:
## Description

It seems algolia docsearch [no longer follows redirects](https://discord.com/channels/1171089640443367494/1171474293398454323/1299061675445583975). This PR adds trailing slashes (intentionally removed by gatsby-config `trailingSlash: 'never'`) to the serialized sitemap.xml URLs so that redirect URLs won't be crawled.

## Detail

While we're here, upgrade `react-components` to v9.1.0 :wink:, which fixes the `Table` API missing `@extends` clause.


ze-flo (approved):
Algolia needs to work on their comms. Great work!

PR #668, opened 2024-11-14
feat: add Color Principles page

ze-flo:
## Description

This PR introduces an initial version of the **Color Principles** page. This page has been generated using the Markdown export feature from Google Docs. Future enhancements and refinements, including custom table designs, will be addressed in subsequent PRs.

## Details
- The existing **Color** page has been renamed to **Palette** and relocated to `design/palette`.
- A new **Color Principles** page has been added at `design/color`.

## Checklist

### DIFF
--- a/src/examples/design/color/BackgroundVariables.tsx
+++ b/src/examples/design/color/BackgroundVariables.tsx
@@ -14,7 +14,7 @@ import {
   IGardenTheme
 } from '@zendeskgarden/react-theming';
 import { Table } from '@zendeskgarden/react-tables';
-import { Code, LG } from '@zendeskgarden/react-typography';
+import { Code, MD } from '@zendeskgarden/react-typography';
 import { Markdown } from 'components/MarkdownProvider/components/Markdown';
 
 const StyledContainer = styled.div`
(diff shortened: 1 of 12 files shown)


lucijanblagonic:
Thanks for pushing the changes. I did a (very quick) first pass. Can we make the following changes:

1. H1: Rename from `Color principles` to `Color` (keep the page name same as the nav item)
2. Introduce H2: `Color principles`

we want to follow this structure:
image

3. Update introduction text from:

> Use this documentation to understand color principles, color palette, color application and theming in Garden.

to (suggestion):

> Color principles, palette, application, light and dark mode considerations in Garden.

^^ There is probably a better description for page level. Suggestions welcome.

4. `Color system / Overview` maybe pick just one, `Overview" (to avoid having "Color" repeating)

5. Table spacing
Would it be possible to add some spacing between heading and a table? Or alternatively a separator 🤔 
image
 
6. Cell spacing
Could we center the swatches within a cell?
image

6. Column widths
Could we adjust column widths to utilize space better? e.g. make swatch widths fixed, as we know what space is needed, and to avoid having description columns break into too many rows?
image

7. Remove icons from anchors
These should all open in the same tab, same as most other links on the website.
image

8. Move this description to `alt`on the image
> Image showing variables attached to a button component. The “foreground.onEmphasis” variable points to the “white” variable indicating the text fill. The “background.primary” variable points to the “blue.700” variable, indicating the background fill.

9. This light / dark mode comparison looks weird because the lack of horizontal space. Compacting the column widths in the first four columns would help a great deal for **Variable** and **Usage** columns content. Alternatively, considering this as an interactive example as @jzempel suggested that changes the swatches/colors names on toggle would also work. 
image

src/examples/design/color/BackgroundVariables.tsx
```diff
@@ -0,0 +1,189 @@
...
+import styled, { useTheme } from 'styled-components';
+import {
+  ColorParameters,
+  DEFAULT_THEME,
+  getColor,
+  IGardenTheme
+} from '@zendeskgarden/react-theming';
+import { Table } from '@zendeskgarden/react-tables';
+import { Code, LG } from '@zendeskgarden/react-typography';
+import { Markdown } from 'components/MarkdownProvider/components/Markdown';
```
jzempel, line 18: I realize these won't compile for CodeSandbox. But I'm not too concerned for a few reasons:
- codesandbox will not be the typical path for these design section examples
- Garden wants to move this `Markdown` component into `@zendeskgarden/react-typography` – at which point we can swap out these imports

smritimadan (approved):
Looks great!

jzempel (approved):

lucijanblagonic (approved):

PR #676, opened 2024-12-16
chore(theming): replace `retrieveComponentStyles` with updated `componentStyles` documentation

jzempel:
## Description

Replace deprecated `retrieveComponentStyles` theming utility documentation with updated `componentStyles`.

## Checklist

- [ ] :memo: ~tested in Chrome, Firefox, Safari, Edge~


src/examples/utilities/components.ts
```diff
@@ -319,22 +334,6 @@ const components = [
...
-        description:
-          "Specifies the unique ID used to retrieve CSS styles from the theme's [components](/components/theme-object#components) object"
-      },
-      props: {
-        required: true,
-        type: 'Partial<ThemeProps<Partial<DefaultTheme>>>',
-        description: 'Provides component props which contain the context `theme` object'
-      }
-    }
-  },
```
ze-flo, line 337: Since v9 still supports `retrieveComponentStyles`, should we mark as deprecated instead of removing the documentation?

jzempel: I considered that, but since it is rarely used, I really want to direct consumers to the right way as the "only" way :wink:

ze-flo (approved):

